PR #2, opened 2021-06-18
Switch to LZO compression for faster startup

AlanGriffiths:


ogra1:
I just did some testing on a Raspberry Pi4 where i re-implemented the splash screen into the boot process ... with the old xz compression ubuntu frame takes about 5-6 seconds to start after the splash is gone from the screen (manually stopwatched), testing with lzo compression got that down to 2-3sec, the difference is significantly noticeable on boot ...

AlanGriffiths:
There's also https://github.com/MirServer/mesa-core20/pull/6

wmww (changes requested):
Looks like ` .idea/` got added before it was put in .gitignore, and needs to be dropped.

Other than that, sounds like a good idea to me.

AlanGriffiths:
> Other than that, sounds like a good idea to me.

Close enough to approval:

bors r=@wmww

PR #15, opened 2021-08-11
Add libfeedback and squeekboard parts

wmww:
I haven't tested installing/running this yet because I'm having problems with snapd system-wide on Arch and didn't fire up Ubuntu yet. It builds on my machine.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -102,6 +102,33 @@ parts:
       - -usr/lib/${SNAPCRAFT_ARCH_TRIPLET}/libdrm.so*
       - -usr/lib/${SNAPCRAFT_ARCH_TRIPLET}/libglapi.so*
 
+  libfeedback:
+    # Required by squeekboard because libfeedback-dev isn't in 20.04
+    plugin: meson
+    source: https://source.puri.sm/Librem5/feedbackd.git
+    source-depth: 1
+    build-packages:
+      - dbus
+      - libgirepository1.0-dev
+      - libglib2.0-dev
+      - libgudev-1.0-dev
+      - libgsound-dev
+      - libjson-glib-dev
+      - valac
+
+  squeekboard:
+    after: [libfeedback]
+    plugin: meson
+    source: https://gitlab.gnome.org/World/Phosh/squeekboard.git
+    source-depth: 1
+    build-packages:
+      - libglib2.0-dev
+      - cargo
+      - wayland-protocols
+      - libgnome-desktop-3-dev
+      - libgtk-3-dev
+      - libwayland-dev
+
   platform:
     after: [ppa-setup]
     plugin: nil


RAOF:
This looks like it should work, but I don't think there's any way to run it as-is? It's not an exposed app, and I don't think ubuntu-frame has any config option to run a client?

AlanGriffiths:
> This looks like it should work, but I don't think there's any way to run it as-is? It's not an exposed app, and I don't think ubuntu-frame has any config option to run a client?

Yep, there's also still stuff to add to Mir. The point is to prove we can build squeekboard for core20 (I failed yesterday).

I suggest we keep this as a feature branch until we've released Mir with the OSK related protocol extensions supported.

AlanGriffiths:
Current status: This successfully creates a squeekboard executable but it doesn't stage all the dependencies:
```plain
$ snap run --shell ubuntu-frame
$ ldd $SNAP/usr/local/bin/squeekboard | grep "not found"
	libgnome-desktop-3.so.19 => not found
	libgtk-3.so.0 => not found
	libgdk-3.so.0 => not found
	libpangocairo-1.0.so.0 => not found
	libpango-1.0.so.0 => not found
	libcairo-gobject.so.2 => not found
	libcairo.so.2 => not found
	libgdk_pixbuf-2.0.so.0 => not found
	libfeedback-0.0.so.0 => not found
```

Saviq:
> Current status: This successfully creates a squeekboard executable but it doesn't stage all the dependencies:
> 
> ```
> $ snap run --shell ubuntu-frame
> $ ldd $SNAP/usr/local/bin/squeekboard | grep "not found"
> 	libgnome-desktop-3.so.19 => not found
> 	libgtk-3.so.0 => not found
> 	libgdk-3.so.0 => not found
> 	libpangocairo-1.0.so.0 => not found
> 	libpango-1.0.so.0 => not found
> 	libcairo-gobject.so.2 => not found
> 	libcairo.so.2 => not found
> 	libgdk_pixbuf-2.0.so.0 => not found
> 	libfeedback-0.0.so.0 => not found
> ```

FWIW this looks like a candidate for the gtk framework snap usage?

AlanGriffiths:
> FWIW this looks like a candidate for the gtk framework snap usage?

Something like `https://snapcraft.io/gnome-3-38-2004`? Yes, that seems to have the necessary stuff. (Note that we likely need to pull it in manually, not with the extension, to use it on non-desktop systems.)

Saviq:
> > FWIW this looks like a candidate for the gtk framework snap usage?
> 
> Something like `https://snapcraft.io/gnome-3-38-2004`?

https://snapcraft.io/docs/gnome-3-34-extension to be exact.

> Yes, that seems to have the necessary stuff. (Note that we likely need to pull it in manually, not with the extension, to use it on non-desktop systems.)

Only if using the extension has negative side effects, IMO.

AlanGriffiths:
> > 
> 
> Only if using the extension has negative side effects, IMO.

My concern is negative side effects: experience shows things like `desktop-launch` are written assuming there is a user session and on core systems either don't work or do unnecessary work for every launch. They are are also fragile on core as they don't get tested for it.

wmww:
I think some of the problems I've been having are because snapd is currently sorta broken on Arch because of cgroup v2 support. Will try tomorrow on Ubuntu.

wmww:
This is now sorta kinda working! All sorts of things are still broken (including positioning), but at least on my machine the OSK now appears.

snap/snapcraft.yaml
```diff
@@ -102,6 +115,59 @@ parts:
...
+    plugin: nil
+    build-packages:
+      - libgdk-pixbuf2.0-0
+      - shared-mime-info
+    stage-packages:
+      - librsvg2-common
+    override-build: |
+      update-mime-database "$SNAPCRAFT_PART_INSTALL/usr/share/mime"
+      mkdir -p "$SNAPCRAFT_PART_INSTALL/etc"
+      "/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/gdk-pixbuf-2.0/gdk-pixbuf-query-loaders" > "$SNAPCRAFT_PART_INSTALL/etc/gdk-pixbuf-loaders.cache"
```
AlanGriffiths, line 169: This looks fragile: You're indexing the loaders in the build environment, not the ones being staged. (But at runtime the index is used with the staged loaders.

wmww:
When I run a GTK app with text input (such as gedit) in this branch of ubuntu-frame Squeekboard automatically appears (tested on Ubuntu 21.04 and Arch).

Current problems I've observed:
- It appears at the top of the screen and doesn't move the other window out of the way
- Bad locale:
```
(process:472618): Gtk-WARNING **: 15:36:18.100: Locale not supported by C library.
	Using the fallback 'C' locale.
```
- No schemas:
```
** (squeekboard:472618): WARNING **: 15:36:18.110: No gsettings schemas installed. Layout switching unavailable.
```
- A few other warnings

wmww:
What does 
```
Errors
------
 - declaration-snap-v2:slots_connection:dbus-session:dbus
	human review required due to 'deny-connection' constraint (interface attributes)
ubuntu-frame_aab99e3-mir2.4.1+dev398_amd64.snap: FAIL
Error: Process completed with exit code 2.
```
mean?

Saviq:
> What does
> ...
> mean?

Early warning about the `dbus` interface being privileged. Fixed in CI by 75dbb72 - but when uploading to store we'll need the store admins to ACK.

AlanGriffiths:
> Early warning about the `dbus` interface being privileged. Fixed in CI by [75dbb72](https://github.com/MirServer/ubuntu-frame/commit/75dbb72869bc74db8affab241cae3a07e865e4b6) - but when uploading to store we'll need the store admins to ACK.

Is a dbus interface going to make sense for a daemon? On Core?

wmww:
Had to make all surfaces maximized instead of fullscreen. This will be fixable once https://github.com/MirServer/mir/pull/2168 lands.

AlanGriffiths:
> Conflicting files
> src/frame_window_manager.cpp

Needs resynchronizing with master

wmww:
I updated this PR to make fullscreen surfaces respect exclusive zones without changing Mir (removing the immediate need for https://github.com/MirServer/mir/pull/2185 or https://github.com/MirServer/mir/pull/2183).

snap/local/slots.json
```diff
@@ -1,5 +1,8 @@
 {
   "wayland":{
     "allow-installation": "true"
+  },
+  "dbus":{
+    "allow-connection": "true"
```
AlanGriffiths, line 6: Shouldn't need this
```suggestion
```

snap/snapcraft.yaml
```diff
@@ -73,7 +80,7 @@ parts:
     plugin: nil
     override-pull: |
       sudo apt --assume-yes install software-properties-common
-      sudo add-apt-repository -y ppa:mir-team/release
+      sudo add-apt-repository -y ppa:mir-team/dev
```
AlanGriffiths, line 83: No longer needed

```suggestion
      sudo add-apt-repository -y ppa:mir-team/release
```

snap/snapcraft.yaml
```diff
@@ -97,6 +104,62 @@ parts:
...
+      - shared-mime-info
+    stage-packages:
+      - librsvg2-common
+    override-build: |
+      update-mime-database ${SNAPCRAFT_PART_INSTALL}/usr/share/mime
+      # Index the pixbuf loaders
+      LOADERS_PATH=$(echo ${SNAPCRAFT_PRIME}/usr/lib/${SNAPCRAFT_ARCH_TRIPLET}/gdk-pixbuf-2.0/*/loaders)
+      QUERY_LOADERS=/usr/lib/${SNAPCRAFT_ARCH_TRIPLET}/gdk-pixbuf-2.0/gdk-pixbuf-query-loaders
+      GDK_PIXBUF_MODULEDIR=${LOADERS_PATH} ${QUERY_LOADERS} > ${LOADERS_PATH}/../loaders.cache
+      sed s!$SNAPCRAFT_PRIME!!g --in-place ${LOADERS_PATH}/../loaders.cache
```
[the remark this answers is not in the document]
AlanGriffiths: You probably had a "dirty" build environment.

wmww:
## Snap size
Uncompressed snaps on my machine:
- current master: 27MB
- this branch (before last commit): 530MB
- this branch (with last commit): 169MB

An improvement to be sure, but this branch still drastically increases the size. The remaining additions can be broken into the following roughly equal pieces, each 20-30MB:
- the squeekboard binary - my local build is ~25% smaller for some reason, idk why, but we're not going to get another drastic reductions
- libicudata - we may or may not be able to get rid of most of this with work
- adwaita icon theme - I have an idea that might be able to strip this down a bit without dropping icons we need
- miscellaneous other libraries - not much we can do about these
- miscellaneous data files (like keyboard symbols and whatnot) - not much we can do about these

I'm going to peruse the content interface approach.

AlanGriffiths:
With this latest iteration squeekboard runs but never appears.
```
** (squeekboard:7808): WARNING **: 12:08:04.225: No system layout present
Debug: Tried file "/home/alan/snap/ubuntu-frame/450/.local/share/squeekboard/keyboards/us.yaml", but it's missing: No such file or directory (os error 2)
Info: Loaded layout Resource: us
Can't connect to the bus: Could not connect: Permission denied. Visibility switching unavailable.
** (squeekboard:7808): WARNING **: 12:08:04.226: Could not connect to session manager: Could not connect: Permission denied
```
Doesn't look fatal, but...

PR #21, opened 2021-09-16
Make all toplevel normal surfaces fullscreen

### LINKED ISSUE #19: fullscreen vs maximize
Right now we set toplevel windows to be fullscreen, but this doesn't propagate to clients. This can be observed by running gedit. In fullscreen mode (press F11) it has no titlebar, but in ubuntu-frame it does have one.

Fixing this is further complicated when introducing the OSK. We want apps to *think* they're fullscreen (maybe?), but be treated by the window manger like they're maximized (so they move out of the way of the OSK). In other words, the exact oposite of the current behavior.

wmww:
Fixes #19. It's unclear to me what this code was supposed to be doing.


AlanGriffiths:
> It's unclear to me what this code was supposed to be doing.

I think it is pretty clear what the code is supposed to do: if the window is hidden, minimized, etc. it doesn't fullscreen it. I still think that reasonable for hidden or minimized windows.

The more "interesting" case is the one you hit is "etc": a maximized window, that is probably better changed to fullscreen.

wmww:
Superseded by #22 and #23

PR #22, opened 2021-09-20
Improve logic for overriding window state

AlanGriffiths:
This was going wrong in several scenarios:

1. Maximized windows were not getting fullscreened
2. Fullscreening a  window (F11 in gedit) led to it being offset from the screen corner

### DIFF
--- a/src/frame_window_manager.cpp
+++ b/src/frame_window_manager.cpp
@@ -30,6 +30,31 @@ namespace ms = mir::scene;
 using namespace miral;
 using namespace miral::toolkit;
 
+namespace
+{
+// When do we override the window state requested by the client?
+bool override(mir::optional_value<MirWindowState> const& requested_state)
+{
+    if (requested_state.is_set())
+    {
+        switch (requested_state.value())
+        {
+        case mir_window_state_minimized:
+        case mir_window_state_hidden:
+        case mir_window_state_attached:
+            return false;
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
I think this fixes #19 but that may need rewording to make that clear

wmww:
This still looks a bit clunky and leaves a lot of (pre-existing) code duplication. I think I can clean it up a bit more.

src/frame_window_manager.cpp
```diff
@@ -30,6 +30,52 @@ namespace ms = mir::scene;
...
+    if (spec.parent().is_set() ? spec.parent().value().lock() : window_info.parent())
+    {
+        return false;
+    }
+
+    // Only override state change if the state is being changed
+    if (!spec.state().is_set())
+    {
+        return false;
+    }
```
Saviq, line 58: Any reason why this wouldn't be the first thing to check?

Saviq (approved):
Just a question about the order in `override_state()`, good otherwise.

bors d+

src/frame_window_manager.cpp
AlanGriffiths: Sure

Saviq:
bors merge

Saviq (approved):
Dude!

bors merge

PR #23, opened 2021-09-21
[Merged by Bors] - Improve logic for overriding window state more

wmww:
Builds on #22. Abstracts window state overriding into a function to reduce code duplication.


AlanGriffiths (approved):
Yes, but now I want to tidy a bit more...

bors merge

PR #24, opened 2021-09-21
Maximize instead of fullscreening windows

wmww:
As discussed in latest sync

### DIFF
--- a/src/frame_window_manager.cpp
+++ b/src/frame_window_manager.cpp
@@ -68,9 +68,9 @@ bool override_state(WindowSpecification& spec, WindowInfo const& window_info)
     default:;
     }
 
-    spec.state() = mir_window_state_fullscreen;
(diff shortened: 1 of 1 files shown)


Saviq:
As it stands, this will result in decorations drawn for CSD, right? I don't think we should merge this before we sort this out.

AlanGriffiths:
> As it stands, this will result in decorations drawn for CSD, right? I don't think we should merge this before we sort that out.

Yup. Been trying this out: a good example is mir-kiosk-kodi which clearly gets confused. (Have yet to check if that supports xdg-decoration.)

We don't see a problem with mir-kiosk as that lets apps "fullscreen" without forcing "maximized". (We could also do this in ubuntu-frame, but that may not be the best answer.)

AlanGriffiths:
We've a better approach in #15 - maybe that could be pulled out?

PR #29, opened 2021-10-13
Respect exclusive zones

wmww:
This augments #28 by preventing the OSK from obscuring the fullscreen app.

### DIFF
--- a/src/frame_window_manager.cpp
+++ b/src/frame_window_manager.cpp
@@ -25,10 +25,99 @@
 
 #include <linux/input.h>
 #include <unistd.h>
+#include <limits>
+#include <iostream>
 
 namespace ms = mir::scene;
 using namespace miral;
 using namespace miral::toolkit;
+
+namespace
+{
+struct WindowData
+{
+    WindowData(bool is_ours)
+        : is_ours{is_ours}
+    {
+    }
+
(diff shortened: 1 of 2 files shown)


src/frame_window_manager.cpp
```diff
@@ -25,23 +25,51 @@
 
 #include <linux/input.h>
 #include <unistd.h>
+#include <limits>
+#include <iostream>
```
AlanGriffiths, line 29: We shouldn't need `<iostream>`

src/frame_window_manager.cpp
```diff
@@ -166,9 +226,143 @@ void FrameWindowManagerPolicy::handle_request_resize(WindowInfo& /*window_info*/
...
+        tools.modify_window(window, spec);
+    }
+}
+
+auto FrameWindowManagerPolicy::container_for_rect(Rectangle const& rect) -> Container&
+{
+    if (containers.empty())
+    {
+        std::cerr << "fatal: container_for_rect() called without any containers" << std::endl;
+        abort();
```
AlanGriffiths, line 313:
```suggestion
        mir::fatal_error_abort("container_for_rect() called without any containers");
```

wmww:
Alternative merged

PR #32, opened 2021-10-19
Force fullsceen windows to respect exclusive zones

AlanGriffiths:
Alternative to #29


AlanGriffiths:
@wmww try again with `snap refresh --channel edge/pr32 ubuntu-frame`

PR #37, opened 2021-10-26
[Merged by Bors] - Authorize ubuntu-frame-osk using apparmor

wmww:
This requires `miral::socket_fd_of()` from miral 3.4, so it switches to the dev Mir PPA and is marked as draft for now.


wmww:
Ah, I didn't realize edge uses the dev Mir ppa

src/frame_main.cpp
```diff
@@ -27,6 +28,10 @@
 #include <miral/set_window_management_policy.h>
 #include <miral/wayland_extensions.h>
 
+#include <unistd.h>
+#include <string.h>
+#include <iostream>
```
AlanGriffiths, line 33: I don't see these being used.

```suggestion
```

src/frame_main.cpp
```diff
@@ -35,6 +40,31 @@ int main(int argc, char const* argv[])
     DisplayConfiguration display_config{runner};
     WaylandExtensions wayland_extensions;
 
+    std::set<std::string> const osk_protocols{
+        WaylandExtensions::zwlr_layer_shell_v1,
+        WaylandExtensions::zwp_virtual_keyboard_v1,
+        WaylandExtensions::zwp_input_method_v2};
+    for (auto const& protocol : osk_protocols)
```
AlanGriffiths, line 47:
```suggestion
        WaylandExtensions::zwp_input_method_v2};

    for (auto const& protocol : osk_protocols)
```

src/snap_name_of.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include <miral/version.h>
+#include <sys/apparmor.h>
+
+auto snap_name_of(miral::Application const& app) -> std::string
+{
+#if MIRAL_VERSION >= MIR_VERSION_NUMBER(3, 4, 0)
+    int const app_fd = miral::socket_fd_of(app);
+#else
+    int const app_fd = -1;
+#endif
```
AlanGriffiths, line 33: When used with Miral < 3.4 this PR should not be changing the existing behaviour. This approach breaks the existing mechanism for enabling `zwlr_layer_shell_v1` etc:  the current demo on https://snapcraft.io/ubuntu-frame-osk doesn't work.

AlanGriffiths (changes requested):
The backwards compatibility needs fixing before we land.

It would be nice to fix https://github.com/MirServer/mir/issues/2194 too.

src/snap_name_of.cpp
```diff
@@ -0,0 +1,70 @@
...
+    if (app_fd < 0)
+    {
+        return "";
+    }
+    else if (aa_getpeercon(app_fd, &label_cstr, &mode_cstr) < 0)
+    {
+        std::cerr
+            << "aa_getpeercon() failed for process " << miral::pid_of(app)
+            << ": " << strerror(errno)
+            << std::endl;
```
AlanGriffiths, line 46:
```suggestion
        mir::log_debug("aa_getpeercon() failed for process %d : %s", miral::pid_of(app), strerror(errno));
```

(This also requires `MIR_LOG_COMPONENT`, which should probably go in `CMakeLists.txt`:
```cmake
target_compile_definitions(frame PRIVATE MIR_LOG_COMPONENT="frame")
```

wmww: Done

src/snap_name_of.cpp
wmww: Fixed

src/frame_main.cpp
wmww: Fixed

src/frame_authorization.cpp
```diff
@@ -0,0 +1,107 @@
...
+        }
+        else
+        {
+            return "";
+        }
+    }
+}
+}
+
+FrameAuthorization::FrameAuthorization(miral::WaylandExtensions& extensions)
```
AlanGriffiths, line 82: Why is this a constructor? The associated class has no state to initialize: everything it does is a side-effect.

wmww: Oh, right, I was going to make it cache the snap names of applications so it didn't have to look them up over and over again. I'll add that momentarily.

AlanGriffiths: In that case there's a risk of a dangling `this` if the client code changes:

```c++
       FrameAuthorization{wayland_extensions}; // Who needs a name?
```

AlanGriffiths: > Oh, right, I was going to make it cache the snap names of applications so it didn't have to look them up over and over again. I'll add that momentarily.

I think the right place to store this data would be `ApplicationInfo::userdata()`, that avoids the need to think about cache invalidation. (C.f. [two hard things](https://www.google.com/search?channel=fs&client=ubuntu&q=two+hard+things))

AlanGriffiths:
@wmww I think you now have everything landed in Mir to rework this with the new 3.4 APIs?

src/frame_authorization.cpp
wmww: Since extensions aren't given an `ApplicationInfo` and plumbing the `WindowManagerTools` needed to get it into the authorizer seems impractical, I left off caching for now.

AlanGriffiths (approved):
WFM

bos r+

PR #41, opened 2021-11-16
[Merged by Bors] - Put layer shell protocol back

wmww:
Fixes a mistake in #40 that slipped through review.


AlanGriffiths (approved):
Oops!

bors r+

PR #43, opened 2021-11-23
Add eglstream-kms

AlanGriffiths:
For testing with Nvidia

### DIFF
--- a/scripts/bin/run-frame
+++ b/scripts/bin/run-frame
@@ -13,4 +13,12 @@ then
   unset WAYLAND_DISPLAY
 fi
 
+# Use hack in snapd for proprietary nVidia drivers mounts the drivers in
+# /var/lib/snapd/lib/gl that needs to be in LD_LIBRARY_PATH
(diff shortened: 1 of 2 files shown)


xnox (approved):

xnox:
I guess this is also in some ways a duplicate of https://github.com/MirServer/ubuntu-frame/pull/44 cause that one would also cause the eglstream platform to be available.

AlanGriffiths:
@xnox we're not going to land this given the current state of Nvidia support on Core.

But there is a corresponding release in the store: `edge/pr43`.

xnox:
> @xnox we're not going to land this given the current state of Nvidia support on Core.
> 
> But there is a corresponding release in the store: `edge/pr43`.

surely it is mostly harmless to land this ;-)

AlanGriffiths:
> surely it is mostly harmless to land this ;-)

You know some folks don't need much reason to think something works before they complain that it doesn't.

Saviq:
This got added through #117

PR #45, opened 2021-12-02
[Merged by Bors] - snapcraft: use package-repositories key

xnox:
Instead of manual ppa-setup, use the now supported for a while
package-repositories key to enable mir-team's ppa.


AlanGriffiths:
Cool. Hadn't noticed that become available.

bors merge

AlanGriffiths (approved):
bors merge

PR #57, opened 2022-03-04
[Merged by Bors] - Add x11-window-title option

graysonguarino:
Adds the x11-window-title option from Mir. Defaults to "Ubuntu Frame".

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -73,6 +73,7 @@ config_entry "console-provider"   "vt"                                  "Require
 config_entry "vt"                 "4"                                   "Virtual terminal to use for display (0 to use current)"
 config_entry "cursor"             "auto"                                "Mouse pointer to use (auto|none|software)"
 config_entry "env-hacks"          "MIR_MESA_KMS_DISABLE_MODESET_PROBE"  "As Frame HAS to run mesa-kms we can safely override the probe for KMS everywhere"
+config_entry "x11-window-title"   "Ubuntu Frame"                        "Default window title when run as Mir on X"
 
 if ! diff "${config_temp}" "${config_file}" > /dev/null; then
   mv "${config_file}"  "${config_file}.save" || true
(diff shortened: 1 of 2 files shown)


snap/snapcraft.yaml
```diff
@@ -8,7 +8,7 @@ license: GPL-3.0
 compression: lzo
 package-repositories:
   - type: apt
-    ppa: mir-team/release
+    ppa: mir-team/dev
```
AlanGriffiths, line 11: We don't want to change this.

snap/hooks/configure
```diff
@@ -73,6 +73,7 @@ config_entry "console-provider"   "vt"                                  "Require
 config_entry "vt"                 "4"                                   "Virtual terminal to use for display (0 to use current)"
 config_entry "cursor"             "auto"                                "Mouse pointer to use (auto|none|software)"
 config_entry "env-hacks"          "MIR_MESA_KMS_DISABLE_MODESET_PROBE"  "As Frame HAS to run mesa-kms we can safely override the probe for KMS everywhere"
+config_entry "x11-window-title"   "Ubuntu Frame"                        "Default window title when run as Mir on X"
```
AlanGriffiths, line 76: The [original issue](https://github.com/MirServer/mir/issues/2219) said "Ubuntu Frame simulator", what is the rationale for changing it?

AlanGriffiths (changes requested):
We definitely don't want to use ppa:mir-team/dev in production. And we need to confirm the title.

snap/hooks/configure
graysonguarino: There is no rationale (other than possibly to match the name of the package). Ubuntu Frame Simulator (capital S preferred?) seems to make more sense in context.

graysonguarino: Possibly Ubuntu Frame (Simulated).

AlanGriffiths: PM is happy with "Ubuntu Frame"

AlanGriffiths (approved):
LGTM

bors merge

PR #58, opened 2022-03-17
[Merged by Bors] - add hardware-observe plug 

ogra1:
add hardware-observe plug to ubuntu-frame.daemon to quieten log spam from libinput when walking /run/udev


AlanGriffiths:
I know it is a rare corner case, but wouldn't this also apply when running under loglnd? (The non-daemon app)

AlanGriffiths (approved):
bors merge

PR #61, opened 2022-04-21
[Merged by Bors] - Add screenshot command

wmww:
Draft because it requires unreleased Mir from the dev PPA.

Used grim (wayshot isn't in the Ubuntu repos, and doesn't seem better in any way relevant to us). Connects to the home plug and saves by default in the current directory.

### DIFF
--- a/scripts/bin/run-client
+++ b/scripts/bin/run-client
@@ -1,0 +1,7 @@
+#!/bin/sh
+set -eu
+
+export XDG_RUNTIME_DIR=$(dirname "$XDG_RUNTIME_DIR")
+export WAYLAND_DISPLAY=$(ls /tmp/frame-display-* | tail -n 1)
+
+exec "$@"
--- a/scripts/bin/run-client
+++ b/scripts/bin/run-client
@@ -2,6 +2,5 @@
 set -eu
 
 export XDG_RUNTIME_DIR=$(dirname "$XDG_RUNTIME_DIR")
-export WAYLAND_DISPLAY=$(ls /tmp/frame-display-* | tail -n 1)
 
 exec "$@"
--- a/scripts/bin/run-client
+++ b/scripts/bin/run-client
@@ -1,6 +1,9 @@
 #!/bin/sh
 set -eu
 
+# This script should be included as part of the command chain so snap run --shell uses it
+
+# Use the user's main XDG runtime dir, instead of the snap-specific one
 export XDG_RUNTIME_DIR=$(dirname "$XDG_RUNTIME_DIR")
 
 exec "$@"
--- a/scripts/bin/screenshot
+++ b/scripts/bin/screenshot
@@ -1,6 +1,8 @@
 #!/bin/sh
 set -eu
 
+SCRIPT_NAME='ubuntu-frame.screenshot'
+
 REQUESTED_OUTPUT=""
 while test $# -gt 0; do
   case $1 in
@@ -11,12 +13,12 @@ while test $# -gt 0; do
       ;;
     -*|--*)
       # We need to catch this or we would accept REQUESTED_OUTPUT values that grim treats as options
-      echo "Unknown option $1" >&2
+      echo "$SCRIPT_NAME: unknown option $1" >&2
       exit 1
       ;;
     *)
       if test ! -z "$REQUESTED_OUTPUT"; then
-        echo "Error: multiple output files specified" >&2
+        echo "$SCRIPT_NAME: please provide a single output file" >&2
         exit 1
       fi
       REQUESTED_OUTPUT="$1"
@@ -49,7 +51,7 @@ done
 # Check if we have permission to write to the file
 if ! touch "$OUTPUT_FILE" 2>/dev/null ; then
     ERROR_MESSAGE=$(touch "$OUTPUT_FILE" 2>&1 | sed 's/.*:\s*//')
-    echo "Can't save to $OUTPUT_FILE: $ERROR_MESSAGE" >&2
+    echo "$SCRIPT_NAME: can't save to $OUTPUT_FILE: $ERROR_MESSAGE" >&2
     exit 1
 fi
 rm "$OUTPUT_FILE"
@@ -57,6 +59,6 @@ rm "$OUTPUT_FILE"
 if $SNAP/usr/bin/grim "$OUTPUT_FILE"; then
   echo "Saved to $OUTPUT_FILE"
 else
-  echo "Failed to save screenshot to $OUTPUT_FILE" >&2
+  echo "$SCRIPT_NAME: failed to save screenshot to $OUTPUT_FILE" >&2
   exit 1
 fi
--- a/scripts/bin/screenshot
+++ b/scripts/bin/screenshot
@@ -47,5 +47,4 @@ while test -z "$OUTPUT_FILE"; do
   fi
 done
 
-#exec $SNAP/usr/bin/grim "$OUTPUT_FILE"
-exec grim "$OUTPUT_FILE"
+exec $SNAP/usr/bin/grim "$OUTPUT_FILE"
--- a/scripts/bin/screenshot
+++ b/scripts/bin/screenshot
@@ -6,7 +6,7 @@ OUTPUT_FILE=""
 while test -z "$OUTPUT_FILE"; do
   # Set the default file name based on the date and time. There is no suffix the first iteration.
   # If that file exists a suffix is added and incremented until a name is found that does not exist.
-  CANDIDATE="$SNAP_USER_COMMON/frame_$(date '+%Y%m%d_%H%M')${SUFFIX:+_$SUFFIX}.png"
+  CANDIDATE="$SNAP_USER_COMMON/frame_$(date --iso-8601=minutes)${SUFFIX:+_$SUFFIX}.png"
   SUFFIX=$(( $SUFFIX + 1 ))
   if test ! -f "$CANDIDATE"; then
     OUTPUT_FILE="$CANDIDATE"
--- a/scripts/bin/screenshot
+++ b/scripts/bin/screenshot
@@ -1,7 +1,7 @@
 #!/bin/sh
 set -eu
 
-OUTPUT_FILE=""
+REQUESTED_OUTPUT=""
 while test $# -gt 0; do
   case $1 in
     -h|--help)
@@ -10,27 +10,36 @@ while test $# -gt 0; do
       exit 0
       ;;
     -*|--*)
-      # We need to catch this or we would accept OUTPUT_FILE values that grim treats as options
+      # We need to catch this or we would accept REQUESTED_OUTPUT values that grim treats as options
       echo "Unknown option $1"
       exit 1
       ;;
     *)
-      if test ! -z "$OUTPUT_FILE"; then
+      if test ! -z "$REQUESTED_OUTPUT"; then
         echo "Error: multiple output files specified"
         exit 1
       fi
-      OUTPUT_FILE="$1"
+      REQUESTED_OUTPUT="$1"
       shift
       ;;
   esac
 done
+
+# OUTPUT_DIR is only used if OUTPUT_FILE is empty
+OUTPUT_DIR="$SNAP_USER_COMMON"
+OUTPUT_FILE=""
+if test -d "$REQUESTED_OUTPUT"; then
+  OUTPUT_DIR="$REQUESTED_OUTPUT"
+else
+  OUTPUT_FILE="$REQUESTED_OUTPUT"
+fi
 
 SUFFIX=""
 # If a file name was specified this is skipped
 while test -z "$OUTPUT_FILE"; do
   # Set the default file name based on the date and time. There is no suffix the first iteration.
   # If that file exists a suffix is added and incremented until a name is found that does not exist.
-  CANDIDATE="$SNAP_USER_COMMON/frame_$(date '+%Y%m%d_%H%M')${SUFFIX:+_$SUFFIX}.png"
+  CANDIDATE="$OUTPUT_DIR/frame_$(date '+%Y%m%d_%H%M')${SUFFIX:+_$SUFFIX}.png"
   SUFFIX=$(( $SUFFIX + 1 ))
   if test ! -f "$CANDIDATE"; then
     OUTPUT_FILE="$CANDIDATE"
--- a/scripts/bin/screenshot
+++ b/scripts/bin/screenshot
@@ -1,0 +1,41 @@
+#!/bin/sh
+set -eu
+
+OUTPUT_FILE=""
+while test $# -gt 0; do
+  case $1 in
+    -h|--help)
+      echo "Take a screenshot"
+      echo "USAGE: ubuntu-frame.screenshot [output-file]"
+      exit 0
+      ;;
+    -*|--*)
+      # We need to catch this or we would accept OUTPUT_FILE values that grim treats as options
+      echo "Unknown option $1"
+      exit 1
+      ;;
+    *)
+      if test ! -z "$OUTPUT_FILE"; then
+        echo "Error: multiple output files specified"
+        exit 1
+      fi
+      OUTPUT_FILE="$1"
+      shift
+      ;;
+  esac
+done
+
+SUFFIX=""
+# If a file name was specified this is skipped
+while test -z "$OUTPUT_FILE"; do
+  # Set the default file name based on the date and time. There is no suffix the first iteration.
+  # If that file exists a suffix is added and incremented until a name is found that does not exist.
+  CANDIDATE="frame_$(date '+%Y%m%d_%H%M')${SUFFIX:+_$SUFFIX}.png"
+  SUFFIX=$(( $SUFFIX + 1 ))
+  if test ! -f "$CANDIDATE"; then
+    OUTPUT_FILE="$CANDIDATE"
+  fi
+done
+
+#exec $SNAP/usr/bin/grim "$OUTPUT_FILE"
+exec grim "$OUTPUT_FILE"
--- a/scripts/bin/screenshot
+++ b/scripts/bin/screenshot
@@ -43,8 +43,12 @@ while test -z "$OUTPUT_FILE"; do
   SUFFIX=$(( $SUFFIX + 1 ))
   if test ! -f "$CANDIDATE"; then
     OUTPUT_FILE="$CANDIDATE"
-    echo "Saving screenshot to $OUTPUT_FILE"
   fi
 done
 
-exec $SNAP/usr/bin/grim "$OUTPUT_FILE"
+if $SNAP/usr/bin/grim "$OUTPUT_FILE"; then
+  echo "Saved to $OUTPUT_FILE"
+else
+  echo "Failed to save screenshot to $OUTPUT_FILE"
+  exit 1
+fi
--- a/scripts/bin/screenshot
+++ b/scripts/bin/screenshot
@@ -47,7 +47,11 @@ while test -z "$OUTPUT_FILE"; do
 done
 
 # Check if we have permission to write to the file
-touch "$OUTPUT_FILE"
+if ! touch "$OUTPUT_FILE" 2>/dev/null ; then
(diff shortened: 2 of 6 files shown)


scripts/bin/screenshot
```diff
@@ -0,0 +1,41 @@
...
+SUFFIX=""
+# If a file name was specified this is skipped
+while test -z "$OUTPUT_FILE"; do
+  # Set the default file name based on the date and time. There is no suffix the first iteration.
+  # If that file exists a suffix is added and incremented until a name is found that does not exist.
+  CANDIDATE="frame_$(date '+%Y%m%d_%H%M')${SUFFIX:+_$SUFFIX}.png"
+  SUFFIX=$(( $SUFFIX + 1 ))
+  if test ! -f "$CANDIDATE"; then
+    OUTPUT_FILE="$CANDIDATE"
+  fi
```
AlanGriffiths, line 37: Doesn't grim already do this by default?

AlanGriffiths:
> Connects to the home plug and saves by default in the current directory

1. Is the home plug useful? On core we'll be running as root.

2. We don't control "the current directory", even with `home` connected it may not be accessible.

Maybe we should place it under somewhere like `$SNAP_DATA_COMMON`?

Saviq:
> > Connects to the home plug and saves by default in the current directory
> 
> 1. Is the home plug useful? On core we'll be running as root.

This sounds like we need to suggest the usage of `sudo` in those circumstances? Since you're likely to _run_ it as a user?

> 2. We don't control "the current directory", even with `home` connected it may not be accessible.
> 
> Maybe we should place it under somewhere like `$SNAP_DATA_COMMON`?

I suppose we could default to that location (and return the saved location) to avoid unpleasantness when it fails to save.

Saviq:
Some experiences to maybe improve:

```shell
# trying as a user when Frame is running as root
user $> ubuntu-frame.screenshot 
ls: cannot access '/tmp/frame-display-wayland-0': Permission denied
failed to create display

# trying without displays
user $> sudo ubuntu-frame.screenshot
no wl_output

# having connected a display
user $> snap restart ubuntu-frame
Restarted.

# trying to save to user's $HOME, despite `ubuntu-frame:home` being connected
user $> sudo ubuntu-frame.screenshot
error while writing to output stream

# trying to save a directory (maybe we could support a target dir, not just a file?)
user $> sudo ubuntu-frame.screenshot snap/ubuntu-frame/common  
error while writing to output stream

# finally
user $> sudo ubuntu-frame.screenshot /var/snap/ubuntu-frame/common/file.png
user $> file /var/snap/ubuntu-frame/common/file.png 
/var/snap/ubuntu-frame/common/file.png: PNG image data, 1920 x 1080, 8-bit/color RGB, non-interlaced
```

src/frame_main.cpp
```diff
@@ -37,6 +49,19 @@ int main(int argc, char const* argv[])
...
+        {
+            auto const target = display_path(runner);
+            auto const link = display_link_path(runner);
+            std::filesystem::create_symlink(target, link);
+        });
+    runner.add_stop_callback([&]
+        {
+            remove(display_link_path(runner).c_str());
+        });
+
```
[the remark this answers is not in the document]
wmww: I'm of the opinion `ubuntu-frame.screenshot` should always take a screenshot of Ubuntu Frame, but we can drop this if you don't think the complexity is worth it.

src/frame_authorization.cpp
```diff
@@ -23,21 +23,27 @@
...
+std::vector<std::pair<std::string, std::vector<std::string>>> const protocols_for_snaps{
+    {"ubuntu-frame-osk", {
+        WaylandExtensions::zwlr_layer_shell_v1,
+        WaylandExtensions::zwp_virtual_keyboard_manager_v1,
+        WaylandExtensions::zwp_input_method_manager_v2,
+    }},
+    {"ubuntu-frame", {
+        WaylandExtensions::zwlr_screencopy_manager_v1,
+    }},
+};
```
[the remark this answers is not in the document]
wmww: I think getting the hard coded refactor in first is a good idea, but yeah long term I could see that making sense.

scripts/bin/screenshot
wmww: It has a version. It's version includes seconds, and overwrites if multiple are taken in the same second, which in practice is about as good. I thought having the name start with `frame_` instead of end with `_grim` would be nice. Another advantage is that rolling our own naming allows us to easily switch screenshot tools.

src/frame_main.cpp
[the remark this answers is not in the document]
AlanGriffiths: We already have to explain the need to set `WAYLAND_DISPLAY` in our tutorials and how-tos, it is simple to apply the same advice to `ubuntu-frame.screenshot`.

In addition, this complexity isn't bullet-proof either: What happens when there are two instances of `ubuntu-frame` running? Or when run as the "wrong" user?

src/frame_authorization.cpp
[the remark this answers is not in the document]
AlanGriffiths: Sure. Maybe it can be `consteval` for now?

wmww:
Constexpr was giving me issues, but I did refactor it such that we can move to making authorization an option easier.

I think I also fixed most of the footguns @Saviq ran into, including saving with an automatic name into a specific directory.

Saviq:
Must say the error messages from grim leave lots to be desired… But short of rolling our own screenshotting tool…

```shell
$ ubuntu-frame.screenshot 
Saving screenshot to /home/michal/snap/ubuntu-frame/common/frame_20220426_1325.png
failed to create display

$ sudo ubuntu-frame.screenshot
Saving screenshot to /root/snap/ubuntu-frame/common/frame_20220426_1325.png
no wl_output

$ sudo ubuntu-frame.screenshot .
Saving screenshot to ./frame_20220426_1341.png
error while writing to output stream
```

I think we should still say "Saving screenshot to…" when given a path:
```shell
$ sudo ubuntu-frame.screenshot /var/snap/ubuntu-frame/common/foo.png
$ sudo ubuntu-frame.screenshot /var/snap/ubuntu-frame/common/foo.png
```

This all looks good:
```shell
$ sudo ubuntu-frame.screenshot
Saving screenshot to /root/snap/ubuntu-frame/common/frame_20220426_1340.png

$ sudo ubuntu-frame.screenshot
Saving screenshot to /root/snap/ubuntu-frame/common/frame_20220426_1340_1.png

$ sudo ubuntu-frame.screenshot /var/snap/ubuntu-frame/common 
Saving screenshot to /var/snap/ubuntu-frame/common/frame_20220426_1341.png
```

AlanGriffiths:
> ```shell
> $ ubuntu-frame.screenshot 
> Saving screenshot to /home/michal/snap/ubuntu-frame/common/frame_20220426_1325.png
> failed to create display
> ```

Maybe we should wait for grim to succeed and report "Screenshot saved to /home/michal/snap/ubuntu-frame/common/frame_20220426_1325.png"?

Saviq:
One last thing that irks me:
```plain
$ ubuntu-frame.screenshot .
touch: cannot touch './frame_20220503_1216.png': Permission denied
```

I know a hack, but I wonder if we should replace `touch` with `$0` here?

And we should work on improving the error messages upstream. If it's trivial to build `grim`, maybe we start with patching it for our use?

wmww:
Maintaining patches is a pain. Detecting the likely error conditions and showing messages from bash is going to be easier.

wmww:
New message:
```
ubuntu-frame.screenshot: can't save to /root/frame_20220503_1049.png: Permission denied
```

scripts/bin/screenshot
```diff
@@ -47,12 +49,16 @@ while test -z "$OUTPUT_FILE"; do
 done
 
 # Check if we have permission to write to the file
-touch "$OUTPUT_FILE"
+if ! touch "$OUTPUT_FILE" 2>/dev/null ; then
+    ERROR_MESSAGE=$(touch "$OUTPUT_FILE" 2>&1 | sed 's/.*:\s*//')
+    echo "$SCRIPT_NAME: can't save to $OUTPUT_FILE: $ERROR_MESSAGE" >&2
```
Saviq, line 54: And maintaining this isn't a pain? ;)

wmww: Not as much as patches, in my opinion

scripts/bin/run-client
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+set -eu
+
+export XDG_RUNTIME_DIR=$(dirname "$XDG_RUNTIME_DIR")
```
Saviq, line 4: Seems excessive to have a separate script just for that? Did you mean it to use it elsewhere, or just future proofing?

Also, a comment why it's necessary to do this (it's not clear to me) could help.

scripts/bin/screenshot
```diff
@@ -0,0 +1,64 @@
+#!/bin/sh
+set -eu
+
+SCRIPT_NAME='ubuntu-frame.screenshot'
```
Saviq, line 4:
```suggestion
SCRIPT_NAME='${SNAP_INSTANCE_NAME}.screenshot'
```

scripts/bin/screenshot
```diff
@@ -0,0 +1,64 @@
...
+
+REQUESTED_OUTPUT=""
+while test $# -gt 0; do
+  case $1 in
+    -h|--help)
+      echo "Take a screenshot"
+      echo "USAGE: ubuntu-frame.screenshot [output-file]"
```
Saviq, line 11:
```suggestion
      echo "USAGE: ubuntu-frame.screenshot [output-file-or-directory]"
```

scripts/bin/screenshot
```diff
@@ -0,0 +1,64 @@
...
+      exit 0
+      ;;
+    -*|--*)
+      # We need to catch this or we would accept REQUESTED_OUTPUT values that grim treats as options
+      echo "$SCRIPT_NAME: unknown option $1" >&2
+      exit 1
+      ;;
+    *)
+      if test ! -z "$REQUESTED_OUTPUT"; then
+        echo "$SCRIPT_NAME: please provide a single output file" >&2
```
Saviq, line 21:
```suggestion
        echo "$SCRIPT_NAME: please provide a single output file or directory" >&2
```

scripts/bin/screenshot
```diff
@@ -0,0 +1,64 @@
...
+        exit 1
+      fi
+      REQUESTED_OUTPUT="$1"
+      shift
+      ;;
+  esac
+done
+
+# OUTPUT_DIR is only used if OUTPUT_FILE is empty
```
Saviq, line 30:
```suggestion
```

scripts/bin/screenshot
Saviq: I would still like us to be good citizens and suggest improvements upstream :)

snap/snapcraft.yaml
```diff
@@ -64,6 +64,15 @@ apps:
...
       - wayland
 
+  screenshot:
+    command-chain: [bin/anon-shm-preload, bin/run-client]
+    command: bin/screenshot
+    plugs:
+      - home
+      - opengl
+    slots:
+      - wayland
```
Saviq, line 74: Hmm? Shouldn't this be up in `plugs`?

snap/snapcraft.yaml
```diff
@@ -64,6 +64,15 @@ apps:
     slots:
       - wayland
 
+  screenshot:
+    command-chain: [bin/anon-shm-preload, bin/run-client]
+    command: bin/screenshot
+    plugs:
+      - home
```
Saviq, line 71:
```suggestion
      - home
      - removable-media
```

Saviq (changes requested):
Some tweaks inline.

scripts/bin/screenshot
wmww: If this was a bug then yes, but I don't see "error while writing to output stream" as all that bad an error message. I get that we want to control what our product spits out so we can make sure our customers understand it, but I don't think it's bothering anyone else too much.

scripts/bin/run-client
wmww: It's so it can be included in the command chain and run as part of `snap run --shell`, comment added.

snap/snapcraft.yaml
wmww: Doesn't work, can't have plug and slot of the same name. This is wrong, but it works.

scripts/bin/screenshot
Saviq: Yeah that's the least problematic one. Except for the "stream" part, which does not ring "file" immediately.

But:
- `failed to create display`: wut? Why are you trying to create a display!?
- `no wl_output`: Gah! What's `wl_output`!?

Maybe it's just me, but I don't think `grim` users should be expected to grok those messages.

snap/snapcraft.yaml
Saviq: Right, but you can be explicit about it:
```yaml
  - wayland-client
    interface: wayland
```

scripts/bin/screenshot
Saviq: Yeah that's the least problematic one. Except for the "stream" part, which does not ring "file" immediately.

But:
- `failed to create display`: wut? Why are you trying to create a display!?
- `no wl_output`: Gah! What's `wl_output`!?

Maybe it's just me, but I don't think `grim` users should be expected to grok those messages.

snap/snapcraft.yaml
wmww: Unfortunately this does not seem to work. I added this under general plugs and `wayland-client` under screenshot app plugs. The snap builds, but it fails to connect to the display.

snap/snapcraft.yaml
wmww: It's not showing up here but I added this

Saviq (approved):
I'm +1 on this one, with the caveat that I'd like to improve the error messages going forward. Preferably through improving them in grim itself.

bors delegate

scripts/bin/screenshot
```diff
@@ -0,0 +1,63 @@
...
+set -eu
+
+SCRIPT_NAME="${SNAP_INSTANCE_NAME}.screenshot"
+
+REQUESTED_OUTPUT=""
+while test $# -gt 0; do
+  case $1 in
+    -h|--help)
+      echo "Take a screenshot"
+      echo "USAGE: $SCRIPT_NAME [output-file-or-directory]"
```
AlanGriffiths, line 11: Given the limitations on saving from a confined snap, and other failure modes, do we really want to enable specifying an output directory? Can't we just save somewhere we know we have permission and report the filename?

scripts/bin/screenshot
```diff
@@ -0,0 +1,63 @@
...
+else
+  OUTPUT_FILE="$REQUESTED_OUTPUT"
+fi
+
+SUFFIX=""
+# If a file name was specified this is skipped
+while test -z "$OUTPUT_FILE"; do
+  # Set the default file name based on the date and time. There is no suffix the first iteration.
+  # If that file exists a suffix is added and incremented until a name is found that does not exist.
+  CANDIDATE="$OUTPUT_DIR/frame_$(date '+%Y%m%d_%H%M')${SUFFIX:+_$SUFFIX}.png"
```
AlanGriffiths, line 43: Maybe stick to ISO formatting? 
```suggestion
  CANDIDATE="$OUTPUT_DIR/frame_$(date '+%Y%m%dT%H:%M:%S')${SUFFIX:+_$SUFFIX}.png"
```

scripts/bin/screenshot
```diff
@@ -0,0 +1,63 @@
...
+# Check if we have permission to write to the file
+if ! touch "$OUTPUT_FILE" 2>/dev/null ; then
+    ERROR_MESSAGE=$(touch "$OUTPUT_FILE" 2>&1 | sed 's/.*:\s*//')
+    echo "$SCRIPT_NAME: can't save to $OUTPUT_FILE: $ERROR_MESSAGE" >&2
+    exit 1
+fi
+rm "$OUTPUT_FILE"
+
+if $SNAP/usr/bin/grim "$OUTPUT_FILE" 2>/dev/null; then
+  echo "Saved to $OUTPUT_FILE"
```
AlanGriffiths, line 59: Should we have (at least an option) to report the unadorned filename. I'm thinking of usage such as a script that does:

```
scp `sudo ubuntu-frame.snapshot` $dest
```

AlanGriffiths (changes requested):
One tweak to naming, one discussion of what we should be trying to support.

snap/snapcraft.yaml
Saviq: This worked for me:

```diff
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -71,15 +71,14 @@ apps:
       - home
       - removable-media
       - opengl
-    slots:
-      # Should be a plug but one snap can't have both the wayland slot and plug. This works.
-      - wayland
 
 plugs:
   graphics-core20:
     interface: content
     target: $SNAP/graphics
     default-provider: mesa-core20
+  wayland-client:
+    interface: wayland
 
 parts:
   recipe-version:
```

There's an extra caveat, you need to manually connect it, which is a reason to just stick to the wayland slot. But then, maybe:

```diff
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -41,13 +41,10 @@ apps:
     command-chain: [bin/run-user, bin/run-frame]
     command: usr/local/bin/frame
     plugs:
-      - opengl                # for compositing
       - login-session-control # to run as a user login (e.g. from a VT)
       - x11                   # to run on a host X11 server
       - network-bind          # to run via X-forwarding (e.g. development in a container/VM)
       - hardware-observe      # libinput likes to scan udev
-    slots:
-      - wayland
 
   daemon:
     command-chain: [bin/run-daemon, bin/run-frame]
@@ -60,9 +57,6 @@ apps:
       HOME: $SNAP_DATA
     plugs:
       - hardware-observe      # libinput likes to scan udev
-      - opengl
-    slots:
-      - wayland
 
   screenshot:
     command-chain: [bin/anon-shm-preload, bin/run-client]
@@ -70,16 +64,16 @@ apps:
     plugs:
       - home
       - removable-media
-      - opengl
-    slots:
-      # Should be a plug but one snap can't have both the wayland slot and plug. This works.
-      - wayland
 
 plugs:
   graphics-core20:
     interface: content
     target: $SNAP/graphics
     default-provider: mesa-core20
+  opengl:
+
+slots:
+  wayland:
 
 parts:
   recipe-version:
```

snap/snapcraft.yaml
```diff
@@ -64,6 +64,17 @@ apps:
...
+      - removable-media
```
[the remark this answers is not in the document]
Saviq: That was a request from me, and I suppose if we drop the destination handling, then those should go away, too.

scripts/bin/screenshot
Saviq: Having gone back'n'forth on this I feel like saving to a custom destination is fragile enough and requires massaging through enough, that I agree it's probably is not worth dealing with the fallout.

If `grim` ever becomes usable on its own for this, without all this cruft around it, we can revisit.

scripts/bin/screenshot
Saviq: With the other changes discussed, we can probably make this the only behaviour.

src/frame_authorization.cpp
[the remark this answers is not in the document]
Saviq: This still needs addressing I believe?

snap/snapcraft.yaml
```diff
@@ -64,6 +64,15 @@ apps:
...
       - wayland
 
+  screenshot:
+    command-chain: [bin/anon-shm-preload, bin/run-client]
+    command: bin/screenshot
+    plugs:
+      - opengl
+    slots:
+      # Should be a plug but one snap can't have both the wayland slot and plug. This works.
+      - wayland
```
AlanGriffiths, line 74: That's what you get for preferring `ubuntu-frame.screenshot` over `ubuntu-frame-screenshot`.

(Sorry, will shut up now)

AlanGriffiths (approved):
WFM

snap/snapcraft.yaml
Saviq: Can be handled separately.

bors merge

PR #69, opened 2022-06-15
Workaround gbm-kms/Nvidia weirdness

AlanGriffiths:
See https://github.com/MirServer/mir/issues/2467


frank-dspeed:
maybe we can simply try them all and return first working? in a serial fashion?

AlanGriffiths:
Not the best way

Issue #72, opened 2022-06-16
Does not pick up "mir-quirks" from a graphics provider until `configure` runs


AlanGriffiths:
I'm [experimenting](https://github.com/MirServer/mesa-core20/pull/17) with providing quirks over this interface and discovered that changes do not take effect unless/until `configure` runs.

We ought to be able to hook the interface (re)connecting...

PR #74, opened 2022-06-16
[Merged by Bors] - Workaround gbm-kms/Nvidia weirdness

AlanGriffiths:
See MirServer/mir#2467

We do this here, not in the graphics-core20 quirks because #72


Saviq (approved):
It's a bit meh that the only way to see this is by looking at the config file…

bors merge

Issue #77, opened 2022-07-21
If aa_getpeercon() fails then ubuntu-frame-* client snaps don't work with default configuration


m2scared:
On some systems (e.g. Archlinux minimal console only installation 5.18.12-arch1-1) this call fails and a workaround is required to enable the extensions needed by, for example, ubuntu-frame-osk.

----

Original report:

Hi,

I have tried to get `ubuntu-frame-osk` to work unsuccessfully by following https://discourse.ubuntu.com/t/ubuntu-frame-osk-documentation/27568.

The service was crashing with `No virtual keyboard manager Wayland global available` on **release** and **edge** versions.

I was only able to get it working when I found this issue https://github.com/MirServer/mir/issues/2063 which resonates with the problem.

By simply adding the config below `ubuntu-frame-osk` works fine on **edge**.

```
snap set ubuntu-frame config "
add-wayland-extensions=zwlr_layer_shell_v1:zwp_virtual_keyboard_manager_v1:zwp_text_input_manager_v3:zwp_input_method_manager_v2
"
```

ubuntu-frame `71-mir2.8.0+dev398`
ubuntu-frame-osk `40-squeekboard-v1.17.1`
wpe-webkit-mir-kiosk `2.30.5`

AlanGriffiths:
Hi, that shouldn't be necessary and I need a bit more information to work out what's happening.

1. What operating system are you using?
2. How are you running Ubuntu Frame and Ubuntu Frame OSK? From the command line? As daemons?

m2scared:
Hi @AlanGriffiths,

1. I am using Archlinux minimal console only installation `5.18.12-arch1-1`.
2. Basically I followed the tutorials for installing Ubuntu Frame and Ubuntu Frame OSK. From command line, yes. Using daemons, yes.

```
git clone https://aur.archlinux.org/snapd.git
cd snapd
makepkg -si
sudo systemctl enable --now snapd.socket
sudo snap install ubuntu-frame --edge
sudo snap set ubuntu-frame daemon=true
sudo snap install wpe-webkit-mir-kiosk
sudo snap connect wpe-webkit-mir-kiosk:wayland
sudo snap set wpe-webkit-mir-kiosk daemon=true
sudo snap set wpe-webkit-mir-kiosk url=https://google.com/
sudo snap start wpe-webkit-mir-kiosk
sudo snap install ubuntu-frame-osk --edge
sudo snap connect ubuntu-frame-osk:wayland
[...]

PR #78, opened 2022-07-25
[Merged by Bors] - Identify client without AppArmor

AlanGriffiths:
Fixes: #77

### DIFF
--- a/src/frame_authorization.cpp
+++ b/src/frame_authorization.cpp
@@ -19,11 +19,13 @@
 #include "frame_authorization.h"
 
 #include <miral/version.h>
+#include <mir/fd.h>
 #include <mir/log.h>
 #include <unistd.h>
 #include <algorithm>
 #include <cstring>
 #include <sys/apparmor.h>
+#include <fcntl.h>
 
 using namespace miral;
 
@@ -55,7 +57,39 @@ auto snap_name_of(miral::Application const& app) -> std::string
     }
     else if (aa_getpeercon(app_fd, &label_cstr, &mode_cstr) < 0)
     {
-        mir::log_debug("aa_getpeercon() failed for process %d: %s", miral::pid_of(app), strerror(errno));
+        mir::log_info("aa_getpeercon() failed for process %d: %s", miral::pid_of(app), strerror(errno));
+
+        if (errno == EINVAL) // EINVAL is what is returned when AppArmor isn't setup
+        {
+            mir::log_info("Fall back (without AppArmor): Identify client via /proc/%%d/cmdline");
+
+            // using the id, find the name of this process
+            char buffer[128] = {'\0' };
+            sprintf(buffer, "/proc/%d/cmdline", miral::pid_of(app));
+
+            if (mir::Fd scoped_fd{open(buffer, O_RDONLY)})
+            {
+                // Read process name from file descriptor scoped_fd
+                // http://linux.die.net/man/2/read
+                if (read(scoped_fd, buffer, sizeof buffer) > 0)
+                {
+                    std::string const path{buffer};
+                    std::string const snap_prefix{"/snap/"};
+
+                    if (path.starts_with(snap_prefix))
+                    {
+                        // Strip the prefix (after_snap_prefix) and app name (before_app_suffix) from the path
+                        auto const after_snap_prefix = begin(path) + snap_prefix.size();
+                        auto const before_app_suffix = std::find(after_snap_prefix, end(path), '/');
+
+                        // We also need to discard any parallel-install suffix (which starts with an underscore)
+                        auto const install_suffix = std::find(after_snap_prefix, before_app_suffix, '_');
+
+                        return std::string{after_snap_prefix, install_suffix};
+                    }
+                }
+            }
+        }
         return "";
     }
     else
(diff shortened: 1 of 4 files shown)


src/frame_authorization.cpp
```diff
@@ -55,7 +57,40 @@ auto snap_name_of(miral::Application const& app) -> std::string
...
+            sprintf(buffer, "/proc/%d/cmdline", miral::pid_of(app));
+            mir::Fd const scoped_fd{open(buffer, O_RDONLY)};
+
+            if (scoped_fd != mir::Fd::invalid)
+            {
+                // Read process name from file descriptor scoped_fd
+                // http://linux.die.net/man/2/read
+                if (read(scoped_fd, buffer, sizeof buffer) > 0)
+                {
+                    std::string const path{buffer};
```
wmww, line 77: `read()` doesn't add a null terminator. `buffer` contains `"/proc/<pid>/cmdline\0"` and is otherwise undefined. Don't we need to use the size returned from `read()` to construct the string?

wmww (changes requested):
I'm generally on-board with the approach, so long as the app armor failure can't be spoofed by a malicious client on a system that does have apparmor support.

I question why we're using the C filesystem functions and strings at all

src/frame_authorization.cpp
RAOF: It is also not an error for `read` to return fewer than the requested number of bytes if interrupted by a signal; this would result in `buffer` not being null-terminated, and an improper memory access in the `string` constructor.

I *think* the *absolutely correct* way to do this would be to:

1. Grab a `pidfd` of the process; I *believe*¹ this ensures the kernel will not recycle the pid, even if the process terminates. (We need this so that we know the `proc` files aren't going to be replaced mid-check)
2. `stat` the `cmdline` file, to determine how many bytes we *expect*
3. `read` from `scoped_fd` until we've read as many bytes as we expect.

I think we can get away with just keeping track of how many bytes `read` has read and null-terminating (or just `memset(buffer, 0, sizeof buffer)`?), since:
1. I don't think it's actually possible to interrupt such a small `read` with a signal part way through, and
2. The failure mode is either a spurious denial or, if the attacker is very lucky, a spurious authorisation (if the attacker has installed `authenticated-snap-but-this-one-is-evil` and manages to interrupt the read at `authenticated-snap`)

For added bonus points we could handle `EINTR` here; interruption by a signal *before* any data is read is much easier to hit. 

¹: But the documentation is not great; this is implied by the docs, not outright stated.

RAOF (changes requested):
As this is a weakening of the security properties (my understanding is that pid racing can be made pretty reliable) I think this might want to be configurable, so that administrators can ensure that their systems don't accidentally fall into a less-secure state.

I think we should definitely address the maybe-`read`-doesn't-end-up-null-terminated issue; I'd be happy with a brute-force `memset` for that, though the fancier solution would also be fine :wink:

AlanGriffiths:
> I think we should definitely address the maybe-`read`-doesn't-end-up-null-terminated issue; I'd be happy with a brute-force `memset` for that, though the fancier solution would also be fine wink

We do already zero initialize the array. (OK, we have overwritten the start, but we are guaranteed a `\0` termination.)

src/frame_authorization.cpp
AlanGriffiths: > `read()` doesn't add a null terminator. `buffer` contains `"/proc/<pid>/cmdline\0"` and is otherwise undefined.

The rest of the array is zero initialised.

RAOF: Hah! I did indeed miss that.

AlanGriffiths:
> Grab a pidfd of the process; I believe¹ this ensures the kernel will not recycle the pid, even if the process terminates. (We need this so that we know the proc files aren't going to be replaced mid-check)

Maybe I'm being thick, but this would all have to happen while maintaining the Wayland socket pair and associated state. Is that event theoretically possible?

AlanGriffiths:
> I question why we're using the C filesystem functions and strings at all

I think your point is proven in that most of the discussion wouldn't happen using c++ file IO

wmww (approved):
LGTM

RAOF:
> > Grab a pidfd of the process; I believe¹ this ensures the kernel will not recycle the pid, even if the process terminates. (We need this so that we know the proc files aren't going to be replaced mid-check)
> 
> Maybe I'm being thick, but this would all have to happen while maintaining the Wayland socket pair and associated state. Is that event theoretically possible?

I think so? Have the malicious client spawn a child process, which initiates the connection, hands the fd to the parent, and then dies. The parent might need to do some libwayland surgery to transfer a bit of state, but this leaves the parent with an open Wayland connection that Mir identifies as coming from a pid that no longer exists.

RAOF (commented):
Yeah, that looks better.

I'd still like this gated behind a config option, or at least have a config option to *dis*able this behaviour.

RAOF (approved):
Yeah, I think I'm OK with this now.

bors r+

AlanGriffiths:
bors retry

AlanGriffiths:
"Component "api.snapcraft.io - refresh" and a few other components are Down" - https://status.snapcraft.io/

Will retry later

AlanGriffiths:
bors retry

Issue #14, opened 2021-07-21
ubuntu-frame may fail to start on hybrid PCs


AlanGriffiths:
Mir's default platform configuration may not work on hybrid systems (depending on the cards available and the outputs attached to them).

Explaining the options in the tutorials seems like overkill, and planned improvements to the automated platform selection should fix this in the foreseeable future.

As an example, to force Ubuntu Frame to ignore `nouveau` and `evdi` drivers add the following lines to the `config` option:
```plain
driver-quirks=skip:driver:nouveau
driver-quirks=skip:driver:evdi
```

PR #79, opened 2022-07-29
[Merged by Bors] - Don't try to use evdi

AlanGriffiths:
Mir platform detection doesn't correctly detect that evdi won't work: Just say "no"

Fixes: #14


Saviq (approved):
Looks legit:

```plain
# Mir's gbm-kms support isn't (yet) working with nvidia
driver-quirks=skip:driver:nvidia
# Mir's gbm-kms support isn't (yet) working with evdi
driver-quirks=skip:driver:evdi
```

bors merge

PR #80, opened 2022-08-02
[Merged by Bors] - Add crash reporter

graysonguarino:
Combines the wallpaper into a CrashReporter which, when a log file is created at a certain location, displays the log file on screen.

Creating as draft because I haven't yet looked into where Snapcraft puts logs by default.


snap/snapcraft.yaml
```diff
@@ -27,6 +27,8 @@ environment:
   __EGL_VENDOR_LIBRARY_DIRS: $SNAP/graphics/glvnd/egl_vendor.d
   # Setup XKB
   XKB_CONFIG_ROOT: $SNAP/usr/share/X11/xkb
+  # Define directory where log files should go
+  LOG_DIR: $SNAP/log
```
Saviq, line 31: I wouldn't call these "logs", as that may suggest we'll be following the log as it gets written to? Maybe something along "diagnostics"?

src/CMakeLists.txt
```diff
@@ -6,19 +6,31 @@ set(CMAKE_CXX_STANDARD 20)
...
-    egwallpaper.cpp egwallpaper.h
     egfullscreenclient.cpp egfullscreenclient.h
+    startup_client.cpp startup_client.h
+)
+
+# Check for boost
+find_package(Boost 1.48.0 COMPONENTS filesystem iostreams REQUIRED)
+include_directories (SYSTEM
+  ${Boost_INCLUDE_DIRS}
 )
```
Saviq, line 23: This should be unnecessary, in modern CMake linking sets up include dirs as well.

src/startup_client.cpp
```diff
@@ -0,0 +1,667 @@
...
+{
+    auto file_observer = FileObserver(log_path);
+    
+    while (true)
+    {
+        if (file_observer.file_updated())
+        {
+            draw_crash_reporter();
+        }
+    }
```
Saviq, line 349: This looks like could be a DoS problem. We should think of only updating it every `$seconds`.

src/startup_client.cpp
```diff
@@ -0,0 +1,667 @@
...
+        BOOST_THROW_EXCEPTION(std::runtime_error("Failed to find a font"));
+    }
+    return path;
+}
+
+FileObserver::FileObserver(Path file_path)
+    : file_path{file_path}
+{
+    // Remove old log (if existant)
+    remove(file_path.c_str());
```
Saviq, line 629: We shouldn't do that, we should use / monitor changes on existing files, too.

Think of a boot script generating the diagnostic file before Frame gets to here.

src/startup_client.cpp
```diff
@@ -0,0 +1,667 @@
...
+        BOOST_THROW_EXCEPTION(std::runtime_error(
+            "Initializing inotify failed with error " + std::to_string(fd)));
+    }
+
+    // Set watch on parent path as watching for file directly
+    // has some complicated behavior
+    wd = inotify_add_watch(
+        fd,
+        file_path.parent_path().c_str(),
+        IN_CREATE | IN_MODIFY
```
Saviq, line 643: Wonder if we should [do `IN_CLOSE_WRITE`](https://stackoverflow.com/questions/32377517/inotify-event-in-modify-occurring-twice-for-tftp-put) instead?

Saviq (changes requested):
> Creating as draft because I haven't yet looked into where Snapcraft puts logs by default.

Snap**craft** doesn't run in this context, FWIW, but regardless - there's nowhere "default" logs are put that you should be considering.

The default path under snap confinement needs to be under `$SNAP_USER_DATA`, as you won't be able to read outside of this location.

In addition, we could think of a [content slot](https://snapcraft.io/docs/content-interface) so that connected confined snaps can write there.

In general, I wouldn't call/think of them as "logs", but rather "diagnostics" - we don't want to be following a log file (not only because that will quickly be unreadable as it fills up the screen, unless we `tail` it by some amount). Operators should craft the diagnostics file according to their needs, so it's mostly static on screen.

snap/snapcraft.yaml
Saviq, line 31: This looks like leftover?

src/startup_client.h
```diff
@@ -0,0 +1,188 @@
...
+        case 2: return r;
+        case 3: return a;
+        default: assert(false);
+        }
+    }
+
+    operator uint32_t() const
+    {
+        return b + (g << 8) + (r << 16) + (a << 24);
+    }
```
AlanGriffiths, line 61: Would probably be wrong if used on big-endian architectures. But as it isn't used, I can't be sure.

Anyway, not needed

src/startup_client.h
```diff
@@ -0,0 +1,188 @@
...
+struct wl_display;
+
+namespace geom = mir::geometry;
+
+struct Pixel
+{
+    uint8_t b;
+    uint8_t g;
+    uint8_t r;
+    uint8_t a;
```
AlanGriffiths, line 43: I don't think alpha is ever usefully anything but 255
```suggestion
    uint8_t const a = 0xff;
```

src/startup_client.h
```diff
@@ -0,0 +1,188 @@
...
+
+    operator uint32_t() const
+    {
+        return b + (g << 8) + (r << 16) + (a << 24);
+    }
+
+    Pixel()
+    {
+        b = g = r = a = 0;
+    }
```
AlanGriffiths, line 66: Use a delegating constructor, not assignment; and, I think, alpha should be 255

src/startup_client.cpp
graysonguarino: Agreed.

src/startup_client.h
graysonguarino: This `Pixel` is designed to maintain compatibility with the old definition of `Pixel` (`typedef uint8_t Pixel`) which meant a "real" pixel (not a subpixel) had to be accessed as a `Pixel*`.

Assigning the alpha value to a const would break the ability to cast a `uint8_t*` pixel to a `Pixel`. Instead, I think we should set the alpha value to `0xFF` via a default in the `Pixel()` constructor.

AlanGriffiths: > ability to cast a uint8_t* pixel to a Pixel

That is undefined behaviour. Best avoided

wmww: C++ can't change the order of data members, so if you static assert that `sizeof(Pixel) == 4` than essentially it is defined that a `Pixel*` can be coasted to a `uint8_t*`, right? How could it represent a `Pixel` as anything other than the 4 `uint8_t`s in a row?

RAOF: It's a type-based aliasing violation; you are only allowed to dereference memory through a type-compatible pointer or an `unsigned char` pointer.

Now, `uint8_t` is *almost certainly* a typedef to `unsigned char` on every platform we care about, but a compiler assuming any Pointer* accesses don't alias an `uint8_t*` accesses would be technically correct.

wmww: [this](https://stackoverflow.com/a/2088428) and [this](https://stackoverflow.com/a/60536461) seem to indicate it's fine in at least some cases. It seems that you have to pay for ?? the actual standard, and the versions I found online aren't very searchable so I can't find what exactly it says, but I know the C standard explicitly allows using a pointer to a struct as a pointer to the first member. Is C++ different?

wmww: [here](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf) is where it is in the C standard, 6.7.2.1 paragraph 13:
> Within a structure object, the non-bit-field members and the units in which bit-fields
reside have addresses that increase in the order in which they are declared. A pointer to a
structure object, suitably con verted, points to its initial member (or if that member is a
bit-field, then to the unit in which it resides), and vice v ersa. There may be unnamed
padding within a structure object, but not at its beginning.

RAOF: Hm. I'm certainly no spec authority, but my reading of that is a guarantee that given
```
struct Foo
{
    int a;
    int b;
};

Foo bar;
```
It's guaranteed that `&bar.a == static_cast<int*>(&bar)`. It doesn't appear to say anything about *access* through that pointer. :woman_shrugging: 

There are then two questions:
1. Is it satisfying the type-based aliasing rules (I'm honestly not sure), and
2. If you want to use the `uint8_t*` as an array, is it guaranteed that `&bar.b == &bar.a + 1` (it is not, compilers are free to add whatever padding they wish).

AlanGriffiths: People, the first member of a structure object can be accessed casting the object address/reference to the member type address/reference. This support is motivated by the "tagged variant" idiom.

But that doesn't imply the opposite conversion, from member address to object address is permitted. That is only guaranteed by the standard _if the original address was taken from an actual object_, and only cast via `void*` or `char*` and back. (One reason for this is that the object type may have stricter alignment constraints than the member type - while this doesn't apply here, nobody was motivated to draft standard text to support such special cases.)

wmww: So what you're saying is
```c++
uint8_t* buffer = new uint8_t[400];
Pixel* pixels = static_cast<Pixel*>(buffer);
```
is undefined but
```c++
Pixel* pixels = new Pixel[100];
uint8_t* buffer = static_cast<uint8_t*>(pixels);
```
is fine?

wmww: > If you want to use the uint8_t* as an array, is it guaranteed that &bar.b == &bar.a + 1 (it is not, compilers are free to add whatever padding they wish).

This is why you static assert that sizeof(Pixel) == 4.

graysonguarino: We only need to assert sizeof(Pixel) == 4 once in the codebase to ensure this since what we're checking is compiler behavior, right? So a `static_assert(sizeof(Pixel) == 4)` after the declaration of `Pixel` would guarantee this?

(I'm currently reading about aliasing [here](https://developers.redhat.com/blog/2020/06/02/the-joys-and-perils-of-c-and-c-aliasing-part-1#objects_and_types_in_c_and_c__).)

AlanGriffiths: > So what you're saying is

No.

The language standard is saying that:

```c++
Pixel foo;

auto bar = reinterpret_cast<uint8_t&>(foo);
```

Is OK (because we're meeting a bunch of criteria, including it being the type of the first member). And that variants that are not equivalent, like accessing the second member, are not defined behaviour.

src/startup_client.h
```diff
@@ -0,0 +1,183 @@
...
+#include <boost/filesystem.hpp>
+
+#include <ft2build.h>
+#include FT_FREETYPE_H
+
+struct wl_display;
+
+namespace geom = mir::geometry;
+
+struct Pixel
```
AlanGriffiths, line 38: Given the discussion this has generated and the problems involved in its use I suspect `Pixel` is a poor abstraction and we would be better without it.

The best way to treat raw memory is as an array of `unsigned char` (with, for the paranoid, an assert that `sizeof(unsigned char) == sizeof(unit8_t)`). IIRC using `unit8_t` is problematic as it doesn't have special treatment in the language, nor is it guaranteed (when it exists) to have the same representation as `unsigned char`. (I agree that it would be mad to use different representations, but I don't think the language standards says anything on the subject.)

src/startup_client.cpp
```diff
@@ -0,0 +1,675 @@
...
+
+    while (true)
+    {
+        if (file_observer.file_updated())
+        {
+            draw();
+        }
+        
+        std::this_thread::sleep_for(std::chrono::seconds(sleep_time));
+    }
```
AlanGriffiths, line 339: Shouldn't there be a way to exit this loop on shutdown? Otherwise the process will "hang".

And shouldn't we wait on ioctl, not burn cycles by sleep/polling?

Have a look into adding an ioctl FD to the `poll()` processing in `FullscreenClient::run()` - that's an efficient way to move processing onto a single thread.

src/startup_client.cpp
```diff
@@ -0,0 +1,675 @@
...
+            "liberation-sans",      // Fedora
+            "liberation",           // Arch
+        }},
+    };
+
+    char const* const font_path_search_paths[]{
+        "/usr/share/fonts/truetype",    // Ubuntu/Debian
+        "/usr/share/fonts/TTF",         // Arch
+        "/usr/share/fonts",             // Fedora/Arch
+    };
```
AlanGriffiths, line 429: We should be supplying the font as part of the snap, no need to be searching for multiple fonts in multiple locations.

src/startup_client.cpp
```diff
@@ -0,0 +1,675 @@
...
+    for (geom::Y buffer_y = buffer_top; buffer_y < buffer_bottom; buffer_y += geom::DeltaY{1})
+    {
+        geom::Y const glyph_y = buffer_y - glyph_offset.dy;
+        uint8_t* const glyph_row = glyph->buffer + (glyph_y.as_int() * glyph->pitch);
+        Pixel* const buffer_row = buf + (buffer_y.as_int() * buf_size.width.as_int());
+
+        for (geom::X buffer_x = buffer_left; buffer_x < buffer_right; buffer_x += geom::DeltaX{1})
+        {
+            geom::X const glyph_x = buffer_x - glyph_offset.dx;
+            uint8_t const glyph_alpha = ((int)glyph_row[glyph_x.as_int()] * colour.a) / 255;
```
AlanGriffiths, line 593: In principle, we don't like C-style casts, but that aside why are we casting to a signed type?

src/startup_client.cpp
```diff
@@ -0,0 +1,675 @@
...
+
+    geom::Y const buffer_top = std::max(top_left.y, geom::Y{});
+    geom::Y const buffer_bottom = std::min(top_left.y + geom::DeltaY{glyph->rows}, as_y(buf_size.height));
+
+    geom::Displacement const glyph_offset = as_displacement(top_left);
+
+    for (geom::Y buffer_y = buffer_top; buffer_y < buffer_bottom; buffer_y += geom::DeltaY{1})
+    {
+        geom::Y const glyph_y = buffer_y - glyph_offset.dy;
+        uint8_t* const glyph_row = glyph->buffer + (glyph_y.as_int() * glyph->pitch);
```
AlanGriffiths, line 587: I think `unit8_t` is technically only right when it is a typedef for `unsigned char`. Better to use `unsigned char`

AlanGriffiths (changes requested):
Sorry, a lot needs dealing with

src/startup_client.cpp
graysonguarino: This will lock us in to the assumption that Frame will only be run as a snap as this can break functionality when running the binary directly. Is this okay?

AlanGriffiths: It locks us into the assumption that the path to the font is supplied. That can be done outside the snap too.

graysonguarino:
Reverting to draft new issues are cleaned up

src/startup_client.h
graysonguarino: I introduced the abstraction to help fix a bug I couldn't figure out, but if it's gonna cause this much trouble, let's shelve it. I got the bug figured out the old way.

src/startup_client.h
```diff
@@ -0,0 +1,183 @@
...
+    void operator()(std::weak_ptr<mir::scene::Session> const& session);
+
+    void stop();
+
+    enum WhichColour{
+        wallpaper_top,
+        wallpaper_bottom,
+        crash_background,
+        crash_text
+    };
```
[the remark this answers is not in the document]
graysonguarino: They say hindsight is 20/20, but I've got no idea why *that* was my solution.

graysonguarino:
Reopening as all necessary changes should be made. Only potential issue I can find is that it appears that the snap is able to see when the diagnostic file is created, but it cannot read its contents (due to permissions issues I assume.) Any advice on fixing this?

src/frame_main.cpp
```diff
@@ -37,21 +37,34 @@ int main(int argc, char const* argv[])
...
                               "wallpaper-top",    "Colour of wallpaper RGB", "0x7f7f7f"},
-            CommandLineOption{[&](auto& option) { wallpaper.bottom(option);},
+            CommandLineOption{[&](auto& option) { startup_client.set_wallpaper_bottom_colour(option);},
                               "wallpaper-bottom", "Colour of wallpaper RGB", "0x1f1f1f"},
+            CommandLineOption{[&](auto& option) { startup_client.set_crash_background_colour(option);},
+                              "crash-background", "Colour of crash screen background RGB", "0x380c24"},
+            CommandLineOption{[&](auto& option) { startup_client.set_crash_text_colour(option);},
+                              "crash-text",       "Colour of crash screen text RGB", "0xffffff"},
+            CommandLineOption{[&] (auto& option) { startup_client.set_diagnostic_path(option);},
+                              "diagnostic-path",  "Path (including filename) of diagnostic file", "/home/$USER/Documents/log/diagnostic.txt"},
```
AlanGriffiths, line 58: This default is _never_ going to work.

Firstly, there's no shell to expand `$USER`, so this will be taken literally;
Secondly, when running as root, `$HOME` is not `'home/root`; and,
Thirdly, when running as a snap, there will be no access to the user's `$HOME` directory.

I suggest:
1. not providing a default (and, if no path is provided, don't try to use it); and,
2. for now, use the `environment` stanza to provide it with:

```yaml
environment:
  # Prep for Mir
  MIR_SERVER_DIAGNOSTIC_PATH: ${SNAP_USER_HOME}/log/diagnostic.txt
```

When doing the content interface, we can redirect at that.

src/egfullscreenclient.cpp
```diff
@@ -133,8 +133,29 @@ void egmde::FullscreenClient::Output::done(void* data, struct wl_output* /*wl_ou
...
+        BOOST_THROW_EXCEPTION(std::runtime_error(
+            "Initializing inotify failed with error " + std::to_string(diagnostic_fd)));
+    }
+
+    return diagnostic_fd;
+}
+
+egmde::FullscreenClient::FullscreenClient(wl_display* display, Path diagnostic_path) :
     flush_signal{::eventfd(0, EFD_SEMAPHORE)},
+    diagnostic_signal{get_diagnostic_fd(diagnostic_path)},
```
AlanGriffiths, line 157: Calling a non-static member function in an initializer list is at best fragile, and very easily leads to undefined behaviour. (It must not access any state that has not already been initialized)

src/startup_client.cpp
```diff
@@ -0,0 +1,527 @@
...
+
+void StartupClient::set_crash_text_colour(std::string const& option)
+{
+    set_colour(option, crash_text_colour);
+}
+
+void StartupClient::set_diagnostic_path(std::string const& option)
+{
+    auto path_string = option;
+    boost::replace_all(path_string, "$USER", getenv("USER"));
```
AlanGriffiths, line 112: Oh, this is a nasty hack to get around the default string having `$USER`. But it will give the impression that shell variables are expanded, and that isn't true in general. (It would be better to do this generating the default value, but I don't recommend having a default value.)

src/startup_client.cpp
```diff
@@ -0,0 +1,527 @@
...
+    auto const path = boost::filesystem::path(path_string);
+    if (boost::filesystem::exists(path.parent_path()))
+    {
+        diagnostic_path = path;
+    }
+    else
+    {
+        BOOST_THROW_EXCEPTION(std::runtime_error(
+            "Log file directory (" + path.parent_path().string() + ") does not exist"));
+    }
```
AlanGriffiths, line 123: This isn't going to work once we move to a content interface: that directory will only exist if the interface is plugged and we don't want to require that.

In any case, Frame should be _able_ to start without the directory existing either because the option isn't supplied and we don't check, or because we simply `log_info` when it doesn't.

I prefer the "no option, don't look" approach. In which case `run-frame` should check for the existence of `$(basename $MIR_SERVER_DIAGNOSTIC_PATH)` and `unset MIR_SERVER_DIAGNOSTIC_PATH` if it isn't there.

Also, any diagnostic should refer to "diagnostic", not to "log".

src/frame_main.cpp
AlanGriffiths: Actually, all that is needed for the content interface is the following:

```yaml
environment:
  ...
  MIR_SERVER_DIAGNOSTIC_PATH: $SNAP/diagnostic/diagnostic.txt
```

and 

```yaml
plugs:
  ...
  ubuntu-frame-diagnostic:
    interface: content
    target: $SNAP/diagnostic
```

AlanGriffiths (changes requested):
The `diagnostic-path` argument should be optional; and, if it is not supplied then, Frame should start without the diagnostic handling.

src/startup_client.cpp
graysonguarino: Agreed, this was the wrong approach.

Now, `set_diagnostic_path()`
1. Checks if $MIR_SERVER_DIAGNOSTIC_PATH exists in the environment, and if it is and no option is given, make that the path.
2. Else, check the validity of the given option (if there is one at all) and either make that the path or disable crash reporting.

src/startup_client.cpp
graysonguarino: This was a temporary hack until I changed this to a regex (forgot to push the `TODO` doc, oops) but now regexing should be entirely unnecessary.

graysonguarino (commented):
I believe that this should now be stable enough for a proper review as a MVP for CrashReporter. This should maintain compatibility both in the snap and when run as the binary, with or without a diagnostic file.

In future PRs (or in this one if requested), it may be wise to detect if a line in the diagnostic is printing offscreen and kick it to a new line.

wmww:
After discussing the current progress in our sync, my recommendation is to get the core crash reporter merged first and then look into adding a delay.

scripts/bin/run-frame
```diff
@@ -13,4 +13,9 @@ then
   unset WAYLAND_DISPLAY
 fi
 
+if [ -z "${MIR_SERVER_DIAGNOSTIC_PATH}"]
+then
+  unset MIR_SERVER_DIAGNOSTIC_PATH
+fi
```
AlanGriffiths, line 19: What is the purpose of this?

"If MIR_SERVER_DIAGNOSTIC_PATH is unset or empty, then unset it"

But we use the `environment` stanza to set it to something.

snap/hooks/install
```diff
@@ -1,6 +1,11 @@
 #!/bin/sh
 set -eux
 
+# Setup diagnostic
+mkdir -p $SNAP_DATA/diagnostic -m 777
+touch $SNAP_DATA/diagnostic/diagnostic
+chmod 666 $SNAP_DATA/diagnostic/diagnostic
```
AlanGriffiths, line 7:
```suggestion
mkdir -p "(dirname "$MIR_SERVER_DIAGNOSTIC_PATH")" -m 777
touch "$MIR_SERVER_DIAGNOSTIC_PATH"
chmod 666 "$MIR_SERVER_DIAGNOSTIC_PATH"
```

src/egfullscreenclient.cpp
```diff
@@ -133,11 +134,30 @@ void egmde::FullscreenClient::Output::done(void* data, struct wl_output* /*wl_ou
...
+            diagnostic_path->parent_path().c_str(),
+            IN_CREATE | IN_CLOSE_WRITE
+        );
+    }
+
+    // Check inotify initializaiton
+    if (diagnostic_signal < 0)
+    {
+        BOOST_THROW_EXCEPTION(std::runtime_error(
+            "Initializing inotify failed with error " + std::to_string(diagnostic_signal)));
```
AlanGriffiths, line 158: There's little point in reporting `-1`, you should be reporting `errno` (and using `std::system_error`).

It also good practice to report `errno` errors as soon as possible after the call that sets it. (And, especially, before using `diagnostic_signal` a few lines above)

src/egfullscreenclient.cpp
```diff
@@ -397,9 +438,16 @@ void egmde::FullscreenClient::run(wl_display* display)
...
             {flush_signal,               POLLIN, 0},
+            {diagnostic_signal,          POLLIN, 0},
             {shutdown_signal,            POLLIN, 0},
         };
 
+    #ifndef NAME_MAX
+    #define NAME_MAX 255
+    #endif
+    static const size_t INOTIFY_BUF_LEN = sizeof(inotify_event) + NAME_MAX + 1;
+    inotify_event inotify_buffer[INOTIFY_BUF_LEN];
```
AlanGriffiths, line 449: I'd be really surprised if what this code does is what you think it does

src/egfullscreenclient.h
```diff
@@ -24,18 +24,26 @@
 
 #include <wayland-client.h>
 
+#include <boost/filesystem.hpp>
+
+#include <sys/inotify.h>
+
 #include <functional>
+#include <ctime>
```
AlanGriffiths, line 32: Is this header needed?

src/startup_client.h
```diff
@@ -0,0 +1,123 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef FRAME_CRASH_REPORTER
+#define FRAME_CRASH_REPORTER
```
AlanGriffiths, line 18: Include guard doesn't match filename

src/startup_client.h
```diff
@@ -0,0 +1,123 @@
...
+
+#include <ft2build.h>
+#include FT_FREETYPE_H
+
+struct wl_display;
+
+namespace geom = mir::geometry;
+using Colour = unsigned char;
+
+class StartupClient
```
AlanGriffiths, line 38: This isn't a good name. It gives no clue what it is for, or that it continues to run after startup.

AlanGriffiths (changes requested):
Lots of nits.

As none of these are deep, maybe we should try pairing to address them? Ping me when online.

src/startup_client.h
graysonguarino: I have decided to go for BackgroundClient as a double-entendre. It 1) displays the wallpaper background and 2) is always running in the background.

snap/snapcraft.yaml
```diff
@@ -26,6 +26,8 @@ environment:
   __EGL_VENDOR_LIBRARY_DIRS: $SNAP/graphics/glvnd/egl_vendor.d
   # Setup XKB
   XKB_CONFIG_ROOT: $SNAP/usr/share/X11/xkb
+  # Setup diagnostic
+  MIR_SERVER_DIAGNOSTIC_PATH: $SNAP_DATA/diagnostic/diagnostic
```
AlanGriffiths, line 30: Just a thought: maybe call it `diagnostic.txt` to give a clue about the content?

src/background_client.cpp
```diff
@@ -0,0 +1,535 @@
...
+{
+    set_colour(option, crash_text_colour);
+}
+
+void BackgroundClient::set_diagnostic_path(std::string const& option)
+{
+    auto option_path = boost::filesystem::path(option);
+
+    // Used as default in snap
+    if (auto env = getenv("MIR_SERVER_DIAGNOSTIC_PATH"))
```
AlanGriffiths, line 111: Don't use `getenv()` for this. Use a `miral::ConfigurationOption`

AlanGriffiths (changes requested):
Didn't finish this review. "Comments so far"

snap/snapcraft.yaml
```diff
@@ -26,6 +26,8 @@ environment:
   __EGL_VENDOR_LIBRARY_DIRS: $SNAP/graphics/glvnd/egl_vendor.d
   # Setup XKB
   XKB_CONFIG_ROOT: $SNAP/usr/share/X11/xkb
+  # Setup diagnostic
+  MIR_SERVER_DIAGNOSTIC_PATH: $SNAP_DATA/diagnostic/diagnostic.txt
```
AlanGriffiths, line 30: Maybe $SNAP_COMMON_DATA would be better?

snap/hooks/install
```diff
@@ -1,6 +1,11 @@
 #!/bin/sh
 set -eux
 
+# Setup diagnostic
+mkdir -p "$(dirname "$MIR_SERVER_DIAGNOSTIC_PATH")" -m 777
+touch "$MIR_SERVER_DIAGNOSTIC_PATH"
```
AlanGriffiths, line 6: Do we actually need to create a file? Should there be some default content?

AlanGriffiths (changes requested):
OK, I tried running this and saw several issues.

1. The text displayed when I first updated `/var/snap/ubuntu-frame/x1/diagnostic/diagnostic.txt` was in the top left corner. It would be better to have a margin or to centre the text (both the miral-shell and egmde examples centre the text);
2. When I subsequently updated the `/var/snap/ubuntu-frame/x1/diagnostic/diagnostic.txt` the text didn't change; and,
3. When I exited Frame there was a `Segmentation fault` (sorry, no dump: it's a snap).

snap/hooks/install
graysonguarino: I believe that the blank file should be created so that clients are guaranteed a file to write to instead of having to create it themselves.

As far as default content goes, since the crash screen won't be triggered until modification of the diagnostic file after polling has started, the default content wouldn't get displayed. I think we should leave the content of the file up to the end user.

graysonguarino:
> OK, I tried running this and saw several issues.
> 
> 1. The text displayed when I first updated `/var/snap/ubuntu-frame/x1/diagnostic/diagnostic.txt` was in the top left corner. It would be better to have a margin or to centre the text (both the miral-shell and egmde examples centre the text);
>> Should this be worried about in the initial PR? I think it would be wise to start a second PR to improve the text handling entirely. For example, this revision will gladly overflow text past the bounds of the display. I think that both of those issues should be addressed later on. Additionally, are we sure we want to center the text? If we are treating the diagnostic file as a "simplified log", shouldn't it read like terminal output? If we are treating this more like a Windows blue screen for example, then the centered text makes more sense. Or should this be a configuration option?
> 2. When I subsequently updated the `/var/snap/ubuntu-frame/x1/diagnostic/diagnostic.txt` the text didn't change; 
>> This was caused by strange behavior when watching the parent directory instead of the diagnostic file. This was initially an intentional design decision because we were expecting the clients to create `diagnostic.txt` once it was ready to display a crash screen, but that expectation has since changed and this behavior is because I hadn't fully updated the code to the new expectation. If we are always guaranteed the existence of `diagnostic.txt`, this simplifies the polling code.
> 3. When I exited Frame there was a `Segmentation fault` (sorry, no dump: it's a snap).
>> This was caused by me holding a `shared_ptr` to `BackgroundClient::Self` instead of a `weak_ptr`. Not entirely sure why I felt that needed to be a `shared_ptr` in the first place.

graysonguarino:
It seems like there was an issue introduced to the snap after d31660e. Not sure why watching the diagnostic file instead of its parent path is messing with inotify. Will look more into this in the morning.

AlanGriffiths:
I'm going to leave a bunch of questions for discussion. @Saviq you probably have thoughts on these:

1. Do we need to provide an option for the font? And do we need to package 20 fonts to support this?
2. Do we want to provide options to control the size of text, and margins? (I had expected to copy the prior art in miral-shell and egmde which auto-scales and centres the text)
3. If the `diagnostic.txt` file is there, do we want to display it (possibly after allowing an interval for clients to start)? If we restart after displaying `diagnostic.txt` then we don't continue to display the `diagnostic.txt`.

Saviq:
>     1. Do we need to provide an option for the font? And do we need to package 20 fonts to support this?

No, just use the Ubuntu font.

>     2. Do we want to provide options to control the size of text, and margins? (I had expected to copy the prior art in miral-shell and egmde which auto-scales and centres the text)

The latter would be best. We then give the operator effectively the same level of control, but not force them to use it (i.e. whatever you do, it will look sane).

>     3. If the `diagnostic.txt` file is there, do we want to display it (possibly after allowing an interval for clients to start)? If we restart after displaying `diagnostic.txt` then we don't continue to display the `diagnostic.txt`.

Yes, we should. The diagnostic file may well be static across the lifetime of the device.

src/background_client.cpp
```diff
@@ -0,0 +1,622 @@
...
+    const std::optional<Path> diagnostic_path;
+
+private:
+    uint font_size;
+
+    uint x_margin_percent;
+    uint y_margin_percent;
+
+    std::mutex mutable buffer_mutex;
+    std::atomic<bool> diag_exists = false;
```
AlanGriffiths, line 127: Dead code

AlanGriffiths (changes requested):
When we start we need to display any existing diagnostics, not wait for the first change.

There are examples of automatically scaling and positioning the text here:

https://github.com/MirServer/mir/blob/d1d67d085c8340999d1ff472113daa93e9436468/examples/example-server-lib/decoration_provider.cpp#L176

And here:

https://github.com/AlanGriffiths/egmde/blob/0f273784f8caddd676fc03e569ff6eaf7416187f/printer.cpp#L93

src/background_client.h
```diff
@@ -0,0 +1,127 @@
...
+
+#include <boost/filesystem.hpp>
+
+#include <ft2build.h>
+#include FT_FREETYPE_H
+
+struct wl_display;
+
+namespace geom = mir::geometry;
+using Colour = unsigned char;
```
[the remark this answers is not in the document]
AlanGriffiths: Here's what I mean:

[diff.txt](https://github.com/MirServer/ubuntu-frame/files/9557710/diff.txt)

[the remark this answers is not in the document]
AlanGriffiths: I'm OK with a typedef (see the diff above). Just not with the way it represents several different things none of which correspond to the name.

snap/snapcraft.yaml
```diff
@@ -103,10 +114,22 @@ parts:
...
       - libapparmor1
+      - libboost-filesystem1.71.0
+      - libboost-iostreams1.71.0
+  
+  assets:
+    plugin: nil
+    stage-packages:
+    - fonts-ubuntu
+    prime:
+    - usr/share/fonts/truetype/ubuntu/Ubuntu-R.ttf
```
AlanGriffiths, line 132: This doesn't quite work as intended:
```plain
$ find /snap/ubuntu-frame/current/usr/share/fonts/
/snap/ubuntu-frame/current/usr/share/fonts/
/snap/ubuntu-frame/current/usr/share/fonts/truetype
/snap/ubuntu-frame/current/usr/share/fonts/truetype/dejavu
/snap/ubuntu-frame/current/usr/share/fonts/truetype/dejavu/DejaVuSans-Bold.ttf
/snap/ubuntu-frame/current/usr/share/fonts/truetype/dejavu/DejaVuSans.ttf
/snap/ubuntu-frame/current/usr/share/fonts/truetype/dejavu/DejaVuSansMono-Bold.ttf
/snap/ubuntu-frame/current/usr/share/fonts/truetype/dejavu/DejaVuSansMono.ttf
/snap/ubuntu-frame/current/usr/share/fonts/truetype/dejavu/DejaVuSerif-Bold.ttf
/snap/ubuntu-frame/current/usr/share/fonts/truetype/dejavu/DejaVuSerif.ttf
/snap/ubuntu-frame/current/usr/share/fonts/truetype/ubuntu
/snap/ubuntu-frame/current/usr/share/fonts/truetype/ubuntu/Ubuntu-R.ttf
```
But poking a bit further, `truetype/dejavu/*` is being pulled in by grim!

AlanGriffiths: I don't think grim needs *any* fonts so we want:
```yaml
  grim:
    plugin: nil
    stage-packages: [grim]
    prime:
      - -usr/share/fonts/truetype/dejavu
```

src/background_client.cpp
```diff
@@ -0,0 +1,578 @@
...
+#include <cstring>
+#include <codecvt>
+#include <sstream>
+#include <thread>
+
+#include <boost/throw_exception.hpp>
+#include <boost/filesystem.hpp>
+#include <boost/algorithm/string/replace.hpp>
+
+#include <miral/configuration_option.h>
```
AlanGriffiths, line 32: We don't use this here!

```suggestion
#include <mir/fatal.h>
```

src/background_client.h
```diff
@@ -0,0 +1,127 @@
...
+
+    Colour wallpaper_top_colour = {127, 127, 127, 255};
+    Colour wallpaper_bottom_colour = {31, 31, 31, 255};
+    Colour crash_background_colour = {36, 12, 56, 255};
+    Colour crash_text_colour = {255, 255, 255, 255};
+
+    uint font_size;
+    
+    uint x_margin_percent;
+    uint y_margin_percent;
```
AlanGriffiths, line 80: I dislike leaving these uninitialized until set by configuration, it takes non-local reasoning to see whether we use uninitialized values.

AlanGriffiths (changes requested):
When we start we need to display any existing diagnostics, not wait for the first change.

There are examples of automatically scaling and positioning the text here:

https://github.com/MirServer/mir/blob/d1d67d085c8340999d1ff472113daa93e9436468/examples/example-server-lib/decoration_provider.cpp#L176

And here:

https://github.com/AlanGriffiths/egmde/blob/0f273784f8caddd676fc03e569ff6eaf7416187f/printer.cpp#L93

src/background_client.h
```diff
@@ -49,15 +49,15 @@ class BackgroundClient
 
     /// Renders background as a gradient from top_colour to bottom_colour
     static void render_background(
-        int32_t width,
-        int32_t height,
+        geom::Width width,
+        geom::Height height,
```
AlanGriffiths, line 53: Why? The change from `int32_t` to `geom::Width` & `geom::Height` makes the code less clear and more verbose.

While the _WholeValue_ pattern can reduce a class of errors, it doesn't help if all that is done is extract the value and use that everywhere.

graysonguarino: I'll talk to Sophie and Chris tonight about the WholeValue pattern because I am clearly a bit foggy on my understanding of it and what problems it solves.

src/frame_main.cpp
```diff
@@ -37,21 +37,32 @@ int main(int argc, char const* argv[])
...
+            ConfigurationOption{[&](auto& option) { background_client.set_crash_background_colour(option);},
+                               "diagnostic-background", "Colour of diagnostic screen background RGB", "0x380c24"},
+            ConfigurationOption{[&](auto& option) { background_client.set_crash_text_colour(option);},
+                               "diagnostic-text",       "Colour of diagnostic screen text RGB", "0xffffff"},
+            ConfigurationOption{[&] (auto& option) { background_client.set_diagnostic_path(option);},
+                               "diagnostic-path",  "Path (including filename) of diagnostic file", ""},
+            ConfigurationOption{[&] (auto& option) { background_client.set_x_margin(option);},
+                               "diagnostic-x-margin",     "Size of x-axis margin (in percentage of display width) for diagnostic", "5"},
+            ConfigurationOption{[&] (auto& option) { background_client.set_y_margin(option);},
+                               "diagnostic-y-margin",     "Size of x-axis margin (in percentage of display height) for diagnostic", "5"},
```
AlanGriffiths, line 62: I see no value in having configuration options for the margin(s). We should simply hard code the value.

(In view of that it is moot that the description is misleading. These values are the total of "top+bottom"/"left+right" margins: i.e. twice size of each of the actual margins.)

AlanGriffiths (changes requested):
Some small nits, but getting close

src/background_client.cpp
```diff
@@ -0,0 +1,608 @@
...
+        {
+            new_pixel[i] = ((current_y * bottom_colour[i] + (height - current_y) * top_colour[i]) / height);
+        }
+        new_pixel[3] = 0xFF;
+
+        // Copy new_pixel to buffer
+        auto const pixel_size = sizeof(new_pixel);
+        for (auto current_x = 0; current_x < pixel_size*width; current_x += pixel_size)
+        {
+            memcpy(buffer + current_x, new_pixel, pixel_size);
```
AlanGriffiths, line 202: Non blocking: you can avoid indexing as follows:
```suggestion
        for (auto current_x = buffer; current_x != buffer + pixel_size*width; current_x += pixel_size)
        {
            memcpy(current_x, new_pixel, pixel_size);
```

But the compiler probably rewrites for you anyway

AlanGriffiths (approved):
I'm OK with this. @graysonguarino might want to tidy up one or more things I mentioned so:

bors delegate

graysonguarino:
bors merge

AlanGriffiths (approved):
bors merge

PR #81, opened 2022-08-02
[Merged by Bors] - snap: bump miral version

Saviq:


snap/snapcraft.yaml
```diff
@@ -106,7 +106,7 @@ parts:
       - libboost1.71-dev
       - libapparmor-dev
     stage-packages:
-      - libmiral4
+      - libmiral5
```
AlanGriffiths, line 109: This will break the build for the beta channel:

```suggestion
      - try:
          - libmiral5
      - else:
          - libmiral4
```

AlanGriffiths (changes requested):

snap/snapcraft.yaml
Saviq: We really need a `stable` branch of some kind…

AlanGriffiths (approved):
WFM

bors merge

PR #83, opened 2022-08-12
[Merged by Bors] - Enable input protocols for VNC snap

wmww:
Possibly now that virtual pointer and seat v8 support have landed in Mir. The `ubuntu-frame-vnc` snap still doesn't seem to be trying input, but it's not obvious where that's disabled.

### DIFF
--- a/src/frame_authorization.cpp
+++ b/src/frame_authorization.cpp
@@ -33,6 +33,8 @@ AuthModel const auth_model{{
     }},
     {"ubuntu-frame-vnc", {
         WaylandExtensions::zwlr_screencopy_manager_v1,
+        WaylandExtensions::zwlr_virtual_pointer_manager_v1,
+        WaylandExtensions::zwp_virtual_keyboard_manager_v1,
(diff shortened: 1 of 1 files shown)


wmww:
Oops, we need https://github.com/MirServer/mir/pull/2553

src/frame_authorization.cpp
```diff
@@ -33,6 +33,8 @@ AuthModel const auth_model{{
     }},
     {"ubuntu-frame-vnc", {
         WaylandExtensions::zwlr_screencopy_manager_v1,
+        WaylandExtensions::zwlr_virtual_pointer_manager_v1,
```
AlanGriffiths, line 36:
```suggestion
#if MIRAL_VERSION >= MIR_VERSION_NUMBER(3, 6, 0)
        WaylandExtensions::zwlr_virtual_pointer_manager_v1,
#endif
```

Saviq (approved):
Works!

bors merge

PR #89, opened 2022-10-05
[Merged by Bors] - Add Frame Diagnostic delay

graysonguarino:
This is dependent on https://github.com/MirServer/mir/pull/2646 but adds the requested delay feature to Frame Diagnostic.

There are currently issues with 2646 that are documented there, so this will start as a draft.


AlanGriffiths (changes requested):
Some quick comments

Saviq (changes requested):
From a user perspective, the diagnostic should only be shown when something went wrong (the client app didn't show up for $delay). So you need to start the alarm on startup, and then when the last surface goes away - and cancel the alarm when a surface comes in.

We will need to think about the wallpaper and the OSK getting in the way of that, though.

Otherwise we'll be showing the diagnostic on e.g. client app getting refreshed.

That said, we may need to come up with some additional logic for when the client app cycles on and off all the time - but that would be a separate task.

src/egfullscreenclient.cpp
```diff
@@ -175,6 +184,53 @@ egmde::FullscreenClient::FullscreenClient(wl_display* display, std::optional<Pat
...
+
+        if (timerfd_settime(timer_fd, 0, &spec, NULL) == -1)
+        {
+            BOOST_THROW_EXCEPTION(std::runtime_error(
+                "Setting diagnostic delay timer failed"));
+        }
+
+        if (runner)
+        {
+            diagnostic_timer_handle = runner->register_fd_handler(mir::Fd{timer_fd}, [this](int){ notify_diagnostic_delay_expired(); });
```
AlanGriffiths, line 224: This will not compile with the current release of Mir (nor the current 2.10 branch)

We can use preprocessor checks of MIRAL_VERSION to conditionally compile code for pre-release versions of Mir.
```c++
#if MIRAL_VERSION >= MIR_VERSION_NUMBER(3, 7, 0)
    // "new Mir" code
#else
    // "old Mir" code
#endif
```

src/frame_main.cpp
```diff
@@ -31,17 +31,17 @@
 int main(int argc, char const* argv[])
 {
     using namespace miral;
-    MirRunner runner{argc, argv};
+    MirRunner* const runner = new MirRunner(argc, argv);
```
AlanGriffiths, line 34: Do not create objects with raw `new` - they are not "owned" and will leak

```suggestion
    MirRunner runner{argc, argv};
```

src/egfullscreenclient.cpp
AlanGriffiths: Presumably on "old" Mir we can just call `notify_diagnostic_delay_expired();` immediately?

AlanGriffiths (changes requested):

AlanGriffiths (approved):
OK, not sure if @Saviq wants to review...

PR #90, opened 2022-10-17
[Merged by Bors] - wm: place windows round-robin on all displays

Saviq:
This is not strictly correct, as overlapping displays and particularly display groups will not get the correct treatment - but is an improvement over the current behaviour.

This requires Mir 2.10 to work.

Also fix copyright and include guard along the way.

### DIFF
--- a/src/frame_window_manager.cpp
+++ b/src/frame_window_manager.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2016-2021 Canonical Ltd.
+ * Copyright © 2016-2022 Canonical Ltd.
  *
  * This program is free software: you can redistribute it and/or modify it
  * under the terms of the GNU General Public License version 2 or 3 as
--- a/src/frame_window_manager.cpp
+++ b/src/frame_window_manager.cpp
@@ -90,6 +90,11 @@ auto FrameWindowManagerPolicy::place_new_window(ApplicationInfo const& app_info,
         WindowInfo window_info{};
         if (override_state(specification, window_info))
         {
+            if (!specification.output_id().is_set() && output_count > 0)
+            {
+                // Place new windows round-robin on all available outputs
+                specification.output_id() = window_count++ % output_count + 1;
+            }
             specification.state() = mir_window_state_maximized;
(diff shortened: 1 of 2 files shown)


Saviq:
This confirms that windows that ask for an output still get it:
frame_2022-10-17T16:32+02:00

And those that don't, get spread out:
frame_2022-10-17T16:30+02:00_1

src/frame_window_manager.cpp
```diff
@@ -90,6 +90,11 @@ auto FrameWindowManagerPolicy::place_new_window(ApplicationInfo const& app_info,
         WindowInfo window_info{};
         if (override_state(specification, window_info))
         {
+            if (!specification.output_id().is_set() && output_count > 0)
+            {
+                // Place new windows round-robin on all available outputs
+                specification.output_id() = window_count++ % output_count + 1;
```
AlanGriffiths, line 96: Readers probably both know their operator precedence well enough to avoid confusion, but it is clearer to use parentheses:

```suggestion
                specification.output_id() = (window_count++ % output_count) + 1;
```

AlanGriffiths (approved):
> This requires Mir 2.10 to work.

What happens if we land it without? It clearly compiles, is it simply non-functional?

> display groups will not get the correct treatment

Is there a usecase that this breaks? Display wall?

Assuming you have reassuring answers...

bors delegate

Saviq:
> > This requires Mir 2.10 to work.
>
> What happens if we land it without? It clearly compiles, is it simply non-functional?

Non-functional indeed.

> > display groups will not get the correct treatment
> 
> Is there a usecase that this breaks? Display wall?

Only if you have more than one group, or a group plus outputs not part of a group, and it won't break more than is broken today - as it will just place as many windows as you have displays in the same display area before it "moves on" to the next display area. frame_2022-10-17T18:24+02:00

Today it will simply never place windows outside of the display area associated with the first output.

> Assuming you have reassuring answers...

bors r=AlanGriffiths

AlanGriffiths (approved):
bors merge

Issue #94, opened 2022-10-21
`--diagnostic-path` crashes when existing on start, but removed


Saviq:
```
$ touch foo
$ ./build-Debug/frame --diagnostic-path $PWD/foo
# ...
# remove the file
# ...
!!! Fatal signal received. Attempting cleanup, but deadlock may occur
Mir fatal error: Unsupported attempt to continue after a fatal signal: SIGSEGV
!!! Fatal signal received. Attempting cleanup, but deadlock may occur
Mir fatal error: Unsupported attempt to continue after a fatal signal: SIGABRT
[1]    661279 IOT instruction (core dumped)  ./build-Debug/frame --diagnostic-path $PWD/foo
```

graysonguarino:
I can't seem to replicate this issue when following these instructions exactly on the newest Mir and Frame builds (building from the `add-crash-reporter-delay` branch). If the issue persists for you on the newest builds, can I get a backtrace?

graysonguarino:
Aha! I think I found what may have caused this. Did you happen to have a directory `$PWD/foo`?

I got a similar crash when I did not specify the file's name and just pointed the diagnostic path at a directory.

AlanGriffiths:
@Saviq can you check that you can reproduce this? Neither @graysonguarino or I can

Saviq:
> @Saviq can you check that you can reproduce this? Neither @graysonguarino or I can

Not any more I can't.

PR #96, opened 2022-10-25
Check if diagnostic file is in relative path after absolute

### LINKED ISSUE #93: `--diagnostic-path` needs to include directory
```
# ...

$ touch foo
# ...

# _now_ it's happy
```

```
                                        file
```

graysonguarino:
~~Fixes #93 by clarifying an absolute path is needed for the `--diagnostic-path` option.~~

~~This is if we consider `./` to be a macro that gets an absolute path from a relative path. Not sure how to specify this without getting wordier.~~

Fixes #93 by first checking if file is in absolute path, then attempts relative path if that fails.

### DIFF
--- a/src/background_client.cpp
+++ b/src/background_client.cpp
@@ -180,7 +180,7 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
         }
 
         BOOST_THROW_EXCEPTION(std::runtime_error(
-            "Diagnostic path (" + option_path.parent_path().string() + ") does not exist"));
+            "Diagnostic path (" + relative_path.parent_path().string() + ") does not exist"));
     }
 }
 
--- a/src/background_client.cpp
+++ b/src/background_client.cpp
@@ -173,7 +173,7 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
     {
         auto const relative_path = boost::filesystem::current_path().append(option);
 
-        if (exists(relative_path))
+        if (exists(relative_path.parent_path()))
         {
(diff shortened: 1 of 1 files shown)


Saviq (changes requested):
As [Alan commented](https://github.com/MirServer/ubuntu-frame/issues/93#issuecomment-1287182956), it's not actually that, rather it needs to resolve a relative path without slashes?

Not sure where this (weird) behaviour comes from…

src/frame_main.cpp
```diff
@@ -55,7 +55,7 @@ int main(int argc, char const* argv[])
             ConfigurationOption{[&](auto& option) { background_client.set_crash_text_colour(option);},
                                "diagnostic-text",       "Colour of diagnostic screen text RGB", "0xffffff"},
             ConfigurationOption{[&] (auto& option) { background_client.set_diagnostic_path(option);},
-                               "diagnostic-path",  "Path (including filename) of diagnostic file", ""},
+                               "diagnostic-path",  "Absolute path (including filename) of diagnostic file", ""},
```
AlanGriffiths, line 58: This is wrong. Relative paths also work

AlanGriffiths (changes requested):
The problem is that `diagnostic-path` is split into "directory" and "file" and then existence of the directory is tested. But, if only a filename is supplied then "directory" is empty and the logic breaks.

graysonguarino:
It may be wise to drop this PR in favor of merging only #97. They both dealt with the same lines of the same code and the newer PR fixes both #93 and #94.

PR #97, opened 2022-10-25
Fix crash from directory passed as diagnostic file

graysonguarino:
This fixes what I believe was causing #94 but is still needed whether or not that was the problem. Frame will crash if a directory is passed as a path ending in a filename to `--diagnostic-path`

### DIFF
--- a/src/background_client.cpp
+++ b/src/background_client.cpp
@@ -165,7 +165,7 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
     }
     
     auto option_path = Path(option);
-    if (exists(option_path.parent_path()))
+    if (exists(option_path.parent_path()) && boost::filesystem::is_regular_file(option_path))
     {
         diagnostic_path = option_path;
     }
@@ -173,7 +173,7 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
     {
         auto const relative_path = boost::filesystem::current_path().append(option);
 
-        if (exists(relative_path.parent_path()))
+        if (exists(relative_path.parent_path()) && boost::filesystem::is_regular_file(relative_path))
         {
             diagnostic_path = relative_path;
             return;
--- a/src/background_client.cpp
+++ b/src/background_client.cpp
@@ -180,7 +180,7 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
         }
 
         BOOST_THROW_EXCEPTION(std::runtime_error(
-            "Diagnostic path (" + option_path.parent_path().string() + ") does not exist"));
+            "Diagnostic path (" + relative_path.parent_path().string() + ") does not exist"));
     }
 }
 
--- a/src/background_client.cpp
+++ b/src/background_client.cpp
@@ -164,16 +164,21 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
         return;
     }
     
-    auto option_path = Path(option);
-    // Check if file's parent dir exists, then check that the file is not a directory
-    if (exists(option_path.parent_path()) && !boost::filesystem::is_directory(option_path))
+    if (option.at(0) == '/')
     {
-        diagnostic_path = option_path;
+        auto const absolute_path = Path(option);
+        if (exists(absolute_path.parent_path()) && !boost::filesystem::is_directory(absolute_path))
+        {
+            diagnostic_path = absolute_path;
+            return;
+        }
+
+        BOOST_THROW_EXCEPTION(std::runtime_error(
+            "Diagnostic path (" + option + ") is invalid"));
     }
     else
     {
         auto const relative_path = boost::filesystem::current_path().append(option);
-
         if (exists(relative_path.parent_path()) && !boost::filesystem::is_directory(relative_path))
         {
             diagnostic_path = relative_path;
--- a/src/background_client.cpp
+++ b/src/background_client.cpp
@@ -173,7 +173,7 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
     {
         auto const relative_path = boost::filesystem::current_path().append(option);
 
-        if (exists(relative_path))
+        if (exists(relative_path.parent_path()))
         {
             diagnostic_path = relative_path;
             return;
--- a/src/background_client.cpp
+++ b/src/background_client.cpp
@@ -165,7 +165,7 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
     }
     
     auto option_path = Path(option);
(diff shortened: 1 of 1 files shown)


src/background_client.cpp
```diff
@@ -165,14 +165,22 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
     }
     
     auto option_path = Path(option);
-    if (exists(option_path.parent_path()))
+    if (exists(option_path.parent_path()) && boost::filesystem::is_regular_file(option_path))
```
wmww, line 168: Why can't `exists(option_path.parent_path())` be dropped? Doesn't `boost::filesystem::is_regular_file(option_path)` return false in all cases where the parent path doesn't exist, making the former call redundant?

graysonguarino: Good point!

src/background_client.cpp
```diff
@@ -165,14 +165,22 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
...
     {
+        auto const relative_path = boost::filesystem::current_path().append(option);
+
+        if (boost::filesystem::is_regular_file(relative_path))
+        {
+            diagnostic_path = relative_path;
+            return;
+        }
+
         BOOST_THROW_EXCEPTION(std::runtime_error(
```
AlanGriffiths, line 182: Probably an issue elsewhere in Frame (so non-blocking), but throwing `mir::AbnormalExit` likely gives better output.

AlanGriffiths (changes requested):
So this now crashes if the file does not exist at startup:

```
$ cmake-build-debug/frame --diagnostic-path foo
...
ERROR: /home/alan/CLionProjects/ubuntu-frame/src/background_client.cpp(182): Throw in function void BackgroundClient::set_diagnostic_path(const string&)
std::exception::what: Diagnostic path (/home/alan/CLionProjects/ubuntu-frame/foo) is invalid
```

```
$ echo plover > foo
$ cmake-build-debug/frame --diagnostic-path foo
...
(normal startup)
```

Vaildiating this requires too much manual testing. We need to automate some tests! (Not necessarily in _this_ PR)

src/background_client.cpp
```diff
@@ -165,14 +165,22 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
     }
     
     auto option_path = Path(option);
-    if (exists(option_path.parent_path()))
+    if (boost::filesystem::is_regular_file(option_path))
     {
         diagnostic_path = option_path;
     }
```
Saviq, line 171: Why doesn't this Just Work™ for relative paths? Does `::is_regular_file` not support relative paths?

Saviq (commented):
I haven't checked yet, but does this still work when the diagnostic file doesn't exist on Frame startup?

src/background_client.cpp
AlanGriffiths: Of course it does

src/background_client.cpp
```diff
@@ -165,22 +165,22 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
...
-        if (exists(relative_path.parent_path()) && boost::filesystem::is_regular_file(relative_path))
+        if (boost::filesystem::is_regular_file(relative_path))
         {
             diagnostic_path = relative_path;
             return;
         }
 
         BOOST_THROW_EXCEPTION(std::runtime_error(
-            "Diagnostic path (" + relative_path.parent_path().string() + ") does not exist"));
+            "Diagnostic path (" + relative_path.string() + ") is invalid"));
```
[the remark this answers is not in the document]
graysonguarino: The relative path diagnostic output is a design choice that may be worth discussing.

Since the user-inputted path gets converted to a relative path "automagically" if an absolute path fails, I figure it may be useful for the end user (say, using `--diagnostic-path=foo.txt` but meaning `--diagnostic-path=/foo.txt`) to see "Frame can't find my diagnostic file at location `/foo.txt` because it is looking in relative path `$PWD/foo.txt`  since I didn't include a slash when putting in the option."

graysonguarino:
> I haven't checked yet, but does this still work when the diagnostic file doesn't exist on Frame startup?

It does now!

AlanGriffiths (changes requested):
```
/cmake-build-debug/frame --diagnostic-path /usr/lib
...
ERROR: /home/alan/CLionProjects/ubuntu-frame/src/background_client.cpp(183): Throw in function void BackgroundClient::set_diagnostic_path(const string&)
std::exception::what: Diagnostic path (/home/alan/CLionProjects/ubuntu-frame/usr/lib) is inval
```

src/background_client.cpp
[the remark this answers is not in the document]
AlanGriffiths: I think you are confused, and risk confusing the user too.

`option_path` may be either relative or absolute and _both_ work fine in the existing code but not with the PR. There is an in the existing code issue (#93) when `option` doesn't contain both a directory and a filename, but that has *nothing* to do with relative or absolute paths. It also has nothing to do with the bug this PR purports to fix.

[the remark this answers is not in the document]
graysonguarino: I think my Git etiquette is to blame here. Both #96 and #97 started to touch the same lines of code and working on both at the same time would lead to two half-complete solutions. Issues #93 and #94 are caused by poor separation of absolute and relative path code paths.

`option_path` is gone now and has been replaced by `absolute_path` in the newest commit. Having separate codepaths for if `--diagnostic-path` is absolute or relative is easier to read and provides more descriptive errors to the user.

Should I close this PR and create a new PR titled "Separate absolute and relative path codepaths for diagnostic-path" which addresses both #93 and #94?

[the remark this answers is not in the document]
AlanGriffiths: > Should I close this PR and create a new PR titled "Separate absolute and relative path codepaths for diagnostic-path" which addresses both #93 and #94?

As you say there are currently two PRs touching this code, both of which mistakenly assume that if the `parent_path()` doesn't exist then the it needs the `current_path()` prepended. That is simply wrong. _There is never a reason to prepend the `current_path()`._

As stated above and in #93 a problem arises when the supplied path doesn't include both a directory component and a filename. I think that is what has started you on the path of prepending `current_path()`, but you are doing that in lots of cases where it isn't appropriate. (Recent examples include when the directory doesn't exist, when the file is a directory, ...)

If you want to close both PRs and submit something that works, please do.

But maybe you first want to script and run some tests of the expected behaviour, that way fixing one problem doesn't create another

graysonguarino:
Closing to create a proper PR which fixes #93 and #94.

PR #98, opened 2022-10-26
[Merged by Bors] - Reset timer when windows closed

### LINKED ISSUE #95: React to clients coming in / away for the diagnostic delay
From a user perspective, the diagnostic should only be shown when something went wrong (the client app didn't show up for $delay). So you need to start the alarm on startup, and then when the last surface goes away - and cancel the alarm when a surface comes in.

We will need to think about the wallpaper and the OSK getting in the way of that, though.

Otherwise we'll be showing the diagnostic on e.g. client app getting refreshed.

That said, we may need to come up with some additional logic for when the client app cycles on and off all the time - but that would be a separate task.

_Originally posted by @Saviq in https://github.com/MirServer/ubuntu-frame/pull/89#pullrequestreview-1151212599_

graysonguarino:
Fixes #95 by making Frame aware of how many windows are opened and closed in the window manager. When a window is opened, the default wallpaper is displayed in the background. When all windows are closed, the diagnostic delay timer is reset.

### DIFF
--- a/src/background_client.cpp
+++ b/src/background_client.cpp
@@ -46,8 +46,9 @@ auto get_font_path() -> Path
 }
 } // namespace
 
-BackgroundClient::BackgroundClient(miral::MirRunner* runner)
-: runner{runner}
+BackgroundClient::BackgroundClient(miral::MirRunner* runner, WindowManagerObserver* window_manager_observer)
+: runner{runner},
+  window_manager_observer{window_manager_observer}
 {
 }
 
@@ -124,11 +125,6 @@ TextRenderer::~TextRenderer()
         mir::log_warning("Failed to uninitialize FreeType with error %d", error);
     }
     library = nullptr;
-}
-
-void BackgroundClient::set_window_manager_observer(WindowManagerObserver* window_manager_observer)
-{
-    this->window_manager_observer = window_manager_observer;
 }
 
 void BackgroundClient::set_colour(std::string const& option, Colour& colour)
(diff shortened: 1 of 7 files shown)


Saviq (commented):
Just one minor inconsistency - if you _don't_ pass a `--diagnostic-delay`, the diagnostic does not display until a client disconnects. When you do provide a delay, it will correctly display after the configured delay after startup.

graysonguarino:
> Just one minor inconsistency - if you _don't_ pass a `--diagnostic-delay`, the diagnostic does not display until a client disconnects. When you do provide a delay, it will correctly display after the configured delay after startup.

Turns out *Frame itself* triggers `FrameWindowManagerPolicy::place_new_window()` on boot, and since it thought that meant a window was open, it did what it was supposed to and cleared the diagnostic!

src/background_client.cpp
```diff
@@ -125,6 +126,11 @@ TextRenderer::~TextRenderer()
     library = nullptr;
 }
 
+void BackgroundClient::set_window_manager_observer(WindowManagerObserver* window_manager_observer)
+{
+    this->window_manager_observer = window_manager_observer;
+}
+
```
AlanGriffiths, line 133: There's no reason for `this->window_manager_observer` to be mutable. There are two better designs that occur immediately:

1. `BackgroundClient` takes a pointer to `WindowManagerObserver` in the constructor
2. `BackgroundClient` implements a `WindowManagerObserver` interface

It would also be possible for `BackgroundClient` to subclass `WindowManagerObserver`, but public inheritance of implementation is a fragile design.

AlanGriffiths (approved):
LGTM

bors merge

PR #100, opened 2022-10-27
Separate paths for relative and absolute diagnostic path

graysonguarino:
Fixes both #93 and #94 by clearly separating the codepaths for absolute and relative path checks on `--diagnostic-path`. The absolute code path is triggered if the path starts with a `/`. Otherwise, it assumes a relative path was provided.


AlanGriffiths:
As you have been told multiple times there is no reason to handle absolute and relative paths differently

PR #103, opened 2022-10-28
Use `boost::filesystem::canonical()` to fix diagnostic path issues

### LINKED ISSUE #93: `--diagnostic-path` needs to include directory

graysonguarino:
Fixes #93 by using `boost::filesystem::canonical()` to convert all inputted paths to absolute paths and deal with most error cases.

It may be wise to reconsider this PR and instead add path checking into `miral::ConfigurationOption`. If/when a client wants to use their own font for the diagnostic screen, we'll be back to doing all this checking internally within Frame when this can all be handled within Miral. I'll propose a PR for the Miral addition today.

### DIFF
--- a/src/background_client.cpp
+++ b/src/background_client.cpp
@@ -175,16 +175,14 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
         return;
     }
     
-    auto option_path = Path(option);
-    if (exists(option_path.parent_path()))
-    {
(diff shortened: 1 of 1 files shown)


graysonguarino:
This solution is incorrect. I had been testing it improperly. `boost::filesystem::absolute()` is the way to go.

PR #104, opened 2022-10-28
[Merged by Bors] - Convert all `--diagnostic-path` paths to absolute

### LINKED ISSUE #93: `--diagnostic-path` needs to include directory

graysonguarino:
Fixes #93 by using `boost::filesystem::absolute()` to convert all inputted paths to absolute paths. Also deals with the error case where a user attempts to input a file as the parent directory.

### DIFF
--- a/src/background_client.cpp
+++ b/src/background_client.cpp
@@ -195,10 +195,10 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
             "Parent of diagnostic path is not a directory." + formatted_path_error));
     }
 
-    if (is_directory(path))
+    if (!is_regular_file(path) && exists(path))
     {
         BOOST_THROW_EXCEPTION(std::runtime_error(
-            "Target of diagnostic path is a directory when it should be a file." + formatted_path_error));
+            "Target of diagnostic path is not a file." + formatted_path_error));
     }
 
     diagnostic_path = path;
--- a/src/background_client.cpp
+++ b/src/background_client.cpp
@@ -195,6 +195,12 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
             "Parent of diagnostic path is not a directory." + formatted_path_error));
     }
 
+    if (!exists(path.parent_path()))
+    {
+        BOOST_THROW_EXCEPTION(std::runtime_error(
+            "Parent of diagnostic path does not exist." + formatted_path_error));
+    }
+
     diagnostic_path = path;
 }
 
--- a/src/background_client.cpp
+++ b/src/background_client.cpp
@@ -189,13 +189,13 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
             "Parent of diagnostic path does not exist." + formatted_path_error));
     }
 
-    if (boost::filesystem::is_regular_file(path.parent_path()))
+    if (is_regular_file(path.parent_path()))
     {
         BOOST_THROW_EXCEPTION(std::runtime_error(
             "Parent of diagnostic path is not a directory." + formatted_path_error));
     }
 
-    if (!boost::filesystem::is_regular_file(path))
+    if (!is_regular_file(path))
     {
         BOOST_THROW_EXCEPTION(std::runtime_error(
             "Target of diagnostic path is a directory when it should be a file." + formatted_path_error));
--- a/src/background_client.cpp
+++ b/src/background_client.cpp
@@ -183,22 +183,22 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
         formatted_path_error.append("\n Resolved path: " + path.string());
     }
 
-    if (!boost::filesystem::is_regular_file(path))
+    if (!exists(path.parent_path()))
     {
         BOOST_THROW_EXCEPTION(std::runtime_error(
-            "Target of diagnostic path is a directory when it should be a file." + formatted_path_error));
+            "Parent of diagnostic path does not exist." + formatted_path_error));
     }
-    
+
     if (boost::filesystem::is_regular_file(path.parent_path()))
     {
         BOOST_THROW_EXCEPTION(std::runtime_error(
             "Parent of diagnostic path is not a directory." + formatted_path_error));
     }
 
-    if (!exists(path.parent_path()))
+    if (!boost::filesystem::is_regular_file(path))
     {
         BOOST_THROW_EXCEPTION(std::runtime_error(
-            "Parent of diagnostic path does not exist." + formatted_path_error));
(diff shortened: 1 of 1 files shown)


src/background_client.cpp
```diff
@@ -175,16 +175,20 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
...
     }
     
-    auto option_path = Path(option);
-    if (exists(option_path.parent_path()))
+    auto const path = boost::filesystem::absolute(option);
+    if (boost::filesystem::is_directory(path))
     {
-        diagnostic_path = option_path;
+        BOOST_THROW_EXCEPTION(std::runtime_error(
+            "Target of diagnostic path (" + path.string() + ") is a directory when it should be a file"));
```
AlanGriffiths, line 182: I think reporting on configuration errors should include the configured value (instead or as well as the path we derive from it)

src/background_client.cpp
```diff
@@ -175,16 +175,20 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
...
     }
-    else
+    
+    if (boost::filesystem::is_regular_file(path.parent_path()))
     {
         BOOST_THROW_EXCEPTION(std::runtime_error(
-            "Diagnostic path (" + option_path.parent_path().string() + ") does not exist"));
+            "Parent of diagnostic path (" + path.parent_path().string() + ") is not a directory"));
```
AlanGriffiths, line 188: I think reporting on configuration errors should include the configured value (instead or as well as the path we derive from it)

AlanGriffiths (changes requested):

AlanGriffiths (changes requested):
```
cmake-build-debug/frame --diagnostic-path nonexisting/tmp
```
Expect: an error
Actual: no error

graysonguarino:
> ```
> cmake-build-debug/frame --diagnostic-path nonexisting/tmp
> ```
> 
> Expect: an error Actual: no error

🤦 I forgot to check that case when migrating from `boost::filesystem::canonical()` to `boost::filesystem::absolute()`.

AlanGriffiths:
> I forgot to check that case

That is the problem with manual testing

src/background_client.cpp
```diff
@@ -175,16 +175,33 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
...
+    {
+        BOOST_THROW_EXCEPTION(std::runtime_error(
+            "Target of diagnostic path is a directory when it should be a file." + formatted_path_error));
+    }
+    
+    if (boost::filesystem::is_regular_file(path.parent_path()))
+    {
+        BOOST_THROW_EXCEPTION(std::runtime_error(
+            "Parent of diagnostic path is not a directory." + formatted_path_error));
+    }
```
AlanGriffiths, line 196: I'm not sure how many types of path are supported by boost.Filesystem, but wouldn't it be clearer to test for `is_directory` when that is the requirement?

src/background_client.cpp
```diff
@@ -175,16 +175,33 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
...
+        formatted_path_error.append("\n Resolved path: " + path.string());
     }
-    else
+
+    if (boost::filesystem::is_directory(path))
+    {
+        BOOST_THROW_EXCEPTION(std::runtime_error(
+            "Target of diagnostic path is a directory when it should be a file." + formatted_path_error));
+    }
+
```
AlanGriffiths, line 191: Looking through `boost::filesystem` it looks like paths can be  `is_regular_file`, `is_directory`, `is_symlink` or `is_other`. If the constraint is that it should be a file, then surely the correct test is `!is_regular_file(path)`?

src/background_client.cpp
```diff
@@ -175,16 +175,33 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
...
+    {
+        BOOST_THROW_EXCEPTION(std::runtime_error(
+            "Parent of diagnostic path does not exist." + formatted_path_error));
+    }
+
+    if (boost::filesystem::is_regular_file(path.parent_path()))
+    {
+        BOOST_THROW_EXCEPTION(std::runtime_error(
+            "Parent of diagnostic path is not a directory." + formatted_path_error));
+    }
```
AlanGriffiths, line 196:
```suggestion
    if (!is_directory_file(path.parent_path()))
    {
        BOOST_THROW_EXCEPTION(std::runtime_error(
            "Parent of diagnostic path is not a directory." + formatted_path_error));
    }
```

src/background_client.cpp
```diff
@@ -175,16 +175,33 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
...
     }
-    else
+
+    if (!exists(path.parent_path()))
+    {
+        BOOST_THROW_EXCEPTION(std::runtime_error(
+            "Parent of diagnostic path does not exist." + formatted_path_error));
+    }
+
+    if (is_regular_file(path.parent_path()))
```
AlanGriffiths, line 192:
```suggestion
    if (!is_directory_file(path.parent_path()))
```

src/background_client.cpp
```diff
@@ -175,16 +175,33 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
...
+        BOOST_THROW_EXCEPTION(std::runtime_error(
+            "Parent of diagnostic path is not a directory." + formatted_path_error));
+    }
+
+    if (!is_regular_file(path))
     {
         BOOST_THROW_EXCEPTION(std::runtime_error(
-            "Diagnostic path (" + option_path.parent_path().string() + ") does not exist"));
+            "Target of diagnostic path is a directory when it should be a file." + formatted_path_error));
     }
```
AlanGriffiths, line 202: The code and the diagnostic disagree

AlanGriffiths (changes requested):
The code and the diagnostics should agree.

See both the latest comment and unresolved comments from before

src/background_client.cpp
```diff
@@ -175,16 +175,33 @@ void BackgroundClient::set_diagnostic_path(std::string const& option)
...
+        BOOST_THROW_EXCEPTION(std::runtime_error(
+            "Parent of diagnostic path is not a directory." + formatted_path_error));
+    }
+
+    if (is_directory(path))
     {
         BOOST_THROW_EXCEPTION(std::runtime_error(
-            "Diagnostic path (" + option_path.parent_path().string() + ") does not exist"));
+            "Target of diagnostic path is a directory when it should be a file." + formatted_path_error));
     }
```
AlanGriffiths, line 202: Doesn't reject `is_symlink` or `is_other`

Issue #109, opened 2022-11-04
diagnostics don't display if OSK is connected


AlanGriffiths:

graysonguarino:
I've spotted a bug in this where closing Firefox with the OSK loaded does not trigger the diagnostic upon Firefox close. Working out why at the moment.

PR #110, opened 2022-11-04
[Merged by Bors] - Don't count OSK as an application

AlanGriffiths:
Fixes: #109

### DIFF
--- a/src/frame_window_manager.cpp
+++ b/src/frame_window_manager.cpp
@@ -77,6 +77,19 @@ bool override_state(WindowSpecification& spec, WindowInfo const& window_info)
 
     return true;
 }
+
+auto is_application(WindowInfo const& window_info)
+{
+    switch (window_info.depth_layer())
+    {
+    case mir_depth_layer_application:
+    case mir_depth_layer_always_on_top:
+        return window_info.state() == mir_window_state_attached;
+
+    default:;
+        return false;
+    }
+}
 }
 
 auto WindowCount::increment_opened() -> unsigned short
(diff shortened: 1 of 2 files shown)


AlanGriffiths:
@Saviq (when you have time) @graysonguarino's tweak works for me, but we need an approver

wmww:
I'd argue that the depth layer is less relevant to if the window should effect the diagnostic, and the real thing you should be looking at is if the window state is attached

AlanGriffiths:
> I'd argue that the depth layer is less relevant to if the window should effect the diagnostic, and the real thing you should be looking at is if the window state is attached

Makes sense (but I'm at EOD)

AlanGriffiths:
Oops! There are scenarios this breaks:

1. start frame & wait for diagnostic
2. start gnome-chess
3. stop gnome-chess

Expect: greyscale (until timeout)
Actual: diagnostic

(Not broken on `main`)

AlanGriffiths:
OK, reworked and working (I think)

src/frame_window_manager.cpp
```diff
@@ -77,6 +77,19 @@ bool override_state(WindowSpecification& spec, WindowInfo const& window_info)
...
+
+auto is_application(WindowInfo const& window_info)
+{
+    switch (window_info.depth_layer())
+    {
+    case mir_depth_layer_application:
+    case mir_depth_layer_always_on_top:
+        return window_info.state() != mir_window_state_attached;
+
+    default:;
```
graysonguarino, line 89: Remove semicolon after `default:`

graysonguarino (approved):
Approving, but with one minor suggestion (unless this is intentional for behavior I don't know about.)

bors delegate

AlanGriffiths:
bors merge

graysonguarino (approved):
bors merge

AlanGriffiths:
bors retry

PR #119, opened 2023-01-05
[Merged by Bors] - Pick up display configuration when run as a user command

AlanGriffiths:
We still won't use the configuration when hosted on X11, but this was causing confusion as:

    sudo ubuntu-frame

behaved differently to

    snap set ubuntu-frame daemon=true


Saviq (approved):
Yup. Works fine.

bors merge

Issue #20, opened 2021-09-14
Difficult to bootstrap the `display` config option


Saviq:
It's currently difficult to retrieve the defaults for `frame.display`. We point at `/var/snap/ubuntu-frame/current/frame.display`, but that's an internal detail.

It would be good to `snapctl set display …` with the default config, since that what ends up in there anyway?

AlanGriffiths:
Once the display layout is set we don't handle subsequent changes to the physical outputs very well. For now, I'd rather we "opt in" by manually setting a configuration. But I'm very open to ideas for improving this.

Saviq:
Well, since the file is written by ubuntu-frame anyway, it's not really opt-in, is it?

AlanGriffiths:
Well, Mir won't write that file until it runs as a daemon for the first time. As an example, I don't have it on my system.

Saviq:
Sure, but on Ubuntu Core, just installing `ubuntu-frame` makes the current setting the persistent one. But it's still not available through `snap get` until you manually `snap set` it.

AlanGriffiths:
Right, but when do you envisage setting it?

There's file until after Mir has run as a daemon for the first time, so dumping `snapctl set display="$(cat ${display_file})"` in `hooks/configure` would not take effect until a subsequent "configure".

We probably want Mir to do some better change management (e.g. adding any new devices it encounters). And maybe have a "dump out the display config and exit" option too.

AlanGriffiths:
Ah! Yeah, it is a bit weird moving it out of the way only in the configure hook.

You'd be happy if we move the `mv "${display_file}"  "${display_file}.save" || true` incantation from `configure` to `run-frame`? Want to PR that?

Saviq:
But isn't `${display_file}` used when you `snap set ...`, too? That doesn't sound right to move it about?

AlanGriffiths:
Yes, if `$(snapctl get display)` returns something (in `configure`) we write the file.

At present, if `$(snapctl get display)` returns nothing (in `configure`) we move the file out of the way.

Hypothetically, "if `$(snapctl get display)` returns nothing (in `run-frame`) we move the file" would give behaviour you're happy with?

PR #122, opened 2023-01-25
[Merged by Bors] - Automate reloading display configuration

AlanGriffiths:
This allows Frame to handle display reconfiguration without a restart.

Features:

1. If there is no `display` option set, then when Frame runs it is set to a default
2. When the `display` option is set:
2.1. With Miral < 3.8 Frame is restarted and will load the new display configuration
2.2. With Miral 3.8 Frame is not restarted and will attempt to reload
3. When the display option is unset, the configuration is cleared and Frame restarted

Note: there is (still) no validation of the configuration being set

Fixes: #20


scripts/bin/daemon-post-stop
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+set -eu
+display_file="$SNAP_DATA/frame.display"
+if [ -z "$(snapctl get display)" ] && [ -e "${display_file}" ]  && ! grep "DO NOT EDIT THIS FILE BY HAND"  "${display_file}"
+then
+  snapctl set display="$(cat "${display_file}")"
+fi
```
Saviq, line 7: I thought of a potential improvement on this? Run Frame in the background, wait a bit for the file to appear, then `exec wait -f` on it…

Saviq: Alternatively, if in snap, we could just read/write directly to snapd?

AlanGriffiths: > we could just read/write directly to snapd

You mean Mir would know which snap configuration the host application wants updated?

AlanGriffiths: In case anything is tracking the pid, we should probably run the "wait on file" in the background and continue to exec Frame

Saviq: > You mean Mir would know which snap configuration the host application wants updated?

Yeah, `snapctl` is just a utility, whether it runs in a hook or not. It should be possible for Frame to just speak to SnapD over its REST API and read/write the display config. Not to mention it could just execute `snapctl`.

scripts/bin/run-daemon
```diff
@@ -1,4 +1,10 @@
 #!/bin/sh
 set -eux
 
+display_file="$SNAP_DATA/frame.display"
+if [ -z "$(snapctl get display)" ] && ! [ -e "${display_file}" ]
+then
+  initialise-display-config&
```
Saviq, line 7: Should we give this a timeout? The file not showing up would likely be a bigger problem, but still? Wouldn't want to run it indefinitely?

scripts/bin/daemon-post-stop
AlanGriffiths: Remember Frame, which reasonably knows the snap option and Mir are separate codebases. I'm questioning whether knowledge of snapd and ubuntu-frame belongs in the Mir codebase

scripts/bin/run-daemon
AlanGriffiths: is there a problem with it waiting indefinitely?

scripts/bin/daemon-post-stop
Saviq: Even so, that's its primary (only?) distribution mechanism, so I wouldn't see a problem with it _detecting_ it's in a snap environment, and reacting appropriately. It could well be optional behaviour, enabled with a command line switch.

scripts/bin/run-daemon
Saviq: I suppose not, if it gets cleaned up with e.g. the snap getting stopped.

snap/hooks/configure
```diff
@@ -21,6 +21,7 @@ EOT
     mv "${display_file}"  "${display_file}.save" || true
     mv "${display_temp}" "${display_file}"
     chmod a+r "${display_file}"
+    ((config_changes+=1))
```
Saviq, line 24: Isn't this missing `#uncomment for…`?

AlanGriffiths: Yes, I missed it being edited when I build the snap. Something clever is needed or it will keep happening

scripts/bin/daemon-post-stop
AlanGriffiths: It isn't the only distribution mechanism for Mir: Mir is in Debian, Fedora, Arch, etc archives. And it is use in Ubuntu Touch (which doesn't yet support snaps).

I think the complexity is better in the snap, than in the Wayland compositor

Saviq: Sure, I didn't mean it should go into Mir, it could live within Frame. Either way, wasn't pushing for this in this PR, just something to consider to simplify the back'n'forth between hooks and wrappers.

Saviq (commented):
Upgrading from `stable` to `edge/pr122` does not populate the `display` variable, because the file doesn't have the `DO NOT EDIT…` header.

(Sh|C)ould we, if the file does exist, `snapctl set display=` its contents on refresh? Or on startup?

AlanGriffiths:
> Upgrading from `stable` to `edge/pr122` does not populate the `display` variable, because the file doesn't have the `DO NOT EDIT…` header.

I suspect the reason is that the file exists (`! [ -e "${display_file}" ]`), not because of the contents

AlanGriffiths:
> (Sh|C)ould we, if the file does exist, `snapctl set display=` its contents on refresh? Or on startup?

No

Saviq:
> I suspect the reason is that the file exists (`! [ -e "${display_file}" ]`), not because of the contents

Well, yes, because Frame from stable created it when it first started. I think if `snapctl get display` is empty, we could read the file in?

It's going to be confusing to have it empty on refresh vs. fresh install, regardless of you not changing anything.

Saviq:
> It's going to be confusing to have it empty on refresh vs. fresh install, regardless of you not changing anything.

That may well be a follow-on, though.

Saviq (approved):
Noice!

Saviq:
bors merge

too

PR #124, opened 2023-01-30
[Merged by Bors] - Use AbnormalExit for bad configuration options

### LINKED ISSUE #99: Throw mir::AbnormalExit if bad configuration options given
Probably an issue elsewhere in Frame (so non-blocking), but throwing `mir::AbnormalExit` likely gives better output.

_Originally posted by @AlanGriffiths in https://github.com/MirServer/ubuntu-frame/pull/97#discussion_r1006081753_

All bad configuration option checks currently throw a Boost `std::runtime_error` exception, where `mir::AbnormalExit` is preferred.

graysonguarino:
Fixes #99. I am curious if this PR should include the same change for each `std::runtime_error` in `background_client.cpp`. Is `mir::AbnormalExit` only for issues upon comeup?

### DIFF
--- a/src/background_client.cpp
+++ b/src/background_client.cpp
@@ -17,6 +17,7 @@
 #include "egfullscreenclient.h"
 #include "background_client.h"
 
+#include "mir/abnormal_exit.h"
 #include "mir/log.h"
 
 #include <chrono>
@@ -177,7 +178,7 @@ void BackgroundClient::set_colour(std::string const& option, Colour& colour)
     }
     else
     {
-        BOOST_THROW_EXCEPTION(std::runtime_error(
+        throw(mir::AbnormalExit(
             "Invalid colour (" + option + ") given in program argument"));
     }
 }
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
> Is `mir::AbnormalExit` only for issues upon comeup?

Mostly yes. It about the intended audience of the logged message. When, for example, the configuration doesn't allow Mir to run we want the message to provide help in correcting it (without raising an issue for us to respond to).

These are "expected" exceptions - handling wrong user input.

```plain
...
Invalid colour (xxx) given in program argument
```
(We could improve that text though)

Other exceptions can occur when something goes wrong during execution. For these the intended audience is us, and we want more diagnostic information:

```plain
...
ERROR: /build/ubuntu-frame/parts/ubuntu-frame/src/background_client.cpp(180): Throw in function void BackgroundClient::set_colour(const string&, unsigned char (&)[4])
std::exception::what: Invalid colour (xxx) given in program argument
```

AlanGriffiths (approved):
bors merge

PR #126, opened 2023-02-13
[Merged by Bors] - Add a display-layout option

AlanGriffiths:
This allows the active display layout to switched without restarting Frame

requires https://github.com/MirServer/mir/pull/2832 to work


wmww (approved):
bors r+

Issue #130, opened 2023-03-10
Latest Version 95-mir2.12.1 released on 8 March 2023. just bricked all my customer tv displays


cluemaster-robert:
The latest version of 95-mir2.12.1 was released on 8 March 2023.

We have been using Ubuntu-Frame for a few years now and everything has been working great. Today all of our customers called our support lines saying that our devices with our TV display software using ubuntu-frame are now booting up and just showing a grey screen forever. Our snap application that has always worked is no longer working at all or starting.  We were able to ssh into our test devices and as soon as we login via ssh, our application would launch and start working. If we reboot the device it would return to the ubuntu-frame grey screen and stay there forever.   We downgraded by using a `snap refresh ubuntu-frame --channel=20/latest`  which downgraded and installed 92-mir2.12.0 and our application immediately started displaying after it was installed and the service started.  Rebooting the PC also was normal and would fully launch the ubuntu-frame and connect our application for video.

Something in yesterday's latest update to Ubuntu-Frame either has a bug or some major change that is causing snaps like ours to no longer work at all.

Saviq:
Hi @cluemaster-robert sorry about that, we must have missed your case in our testing (of which there's plenty, I guarantee you).

First, you can [`snap revert <snap>`](https://snapcraft.io/docs/getting-started#heading--revert) any time you encounter any problem with an upgrade - it will go back to the previous revision.

I've now rolled back the release, so a manual `snap refresh`, or a reboot or a wait until the next automatic refresh will bring the previous revision in. That should bring your customers back (and not hurt any more, if they didn't yet receive the update).

Now. To help debug what's gone wrong, may I ask that you attach the logs from both Ubuntu Frame and your application (e.g. [`snap logs -n500 <snap>`](https://snapcraft.io/docs/service-management#heading--logs)).

You should subscribe to [the Mir discourse category](https://discourse.ubuntu.com/c/mir/15) in which we announce a call for testing for each Frame release (e.g. [the last one](https://discourse.ubuntu.com/t/call-for-testing-mir-kiosk-mir-test-tools-egmde-ubuntu-frame-ubuntu-frame-vnc/34270)). You also should run your testing devices on the `candidate` channel, in which we keep our next stable for at least a week during the call for testing.

When we understand your issue, we'll try and include your case in our testing. If you want official support and our QA pipeline replicate your exact setup, please contact our support team at https://ubuntu.com/support.

AlanGriffiths:
[...]

PR #131, opened 2023-03-10
[Merged by Bors] - Unblock clients with a racy `wayland-launch` script

AlanGriffiths:
Fixes: #130

### DIFF
--- a/scripts/bin/run-daemon
+++ b/scripts/bin/run-daemon
@@ -6,4 +6,7 @@ then
   initialise-display-config&
 fi
 
+# Unblock racy clients that hang https://github.com/MirServer/ubuntu-frame/issues/130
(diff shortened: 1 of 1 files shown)


scripts/bin/run-daemon
```diff
@@ -1,10 +1,12 @@
 #!/bin/sh
```
Saviq, line 1:
```suggestion
#!/bin/bash
```

Otherwise:

```plain
2023-03-10T17:03:34+01:00 ubuntu-frame.daemon[2888442]: /snap/ubuntu-frame/5054/bin/run-daemon: 10: disown: not found
2023-03-10T17:03:34+01:00 ubuntu-frame.daemon[2888475]: + sh -c sleep 5; mkdir "/run/user/007"; rmdir "/run/user/007"
```

Saviq (changes requested):
Bug

PR #138, opened 2023-04-19
snap: add missing freetype6

Saviq:


Saviq:
```
# it's here on x86
$ ls -l /snap/core22/current/usr/lib/*/libfreetype.so.6
lrwxrwxrwx 1 root root 21 lip 19  2022 /snap/core22/current/usr/lib/x86_64-linux-gnu/libfreetype.so.6 -> libfreetype.so.6.18.1

# but not on ARM
$ ls -l /snap/core22/current/usr/lib/*/libfreetype.so.6
ls: cannot access '/snap/core22/current/usr/lib/*/libfreetype.so.6': No such file or directory
```

Saviq:
D'uh. Had an outdated core22.

PR #141, opened 2023-06-07
snap: drop apparmor

Saviq:
It's part of the core22 snap.


AlanGriffiths (approved):
Yeah, it's being stripped by snapcraft anyway

PR #144, opened 2023-08-11
snap: add chvt and the session wrapper

Saviq:


Saviq:
Found a way to do this without `chvt`, which means I can keep the session wrapper in the gadget. \o\

PR #145, opened 2023-08-22
snap: bump to libmiral6

Saviq:


AlanGriffiths (approved):
Yes, least bad option

PR #152, opened 2023-10-02
Draw the background/diagnostic on the worker thread

### LINKED ISSUE #151: 50% success rate of the app starting
Using the following:

```
ubuntu-frame       113-mir2.15.0+dev124    6905   22/edge/…      canonical✓  -
ubuntu-frame-osk   44-squeekboard-v1.17.1  316    latest/stable  canonical✓  -
```

This is maybe a little bit better, but still the same issue.

It's about a 50% success rate of the app loading and going full screen or attempting to load, failing a bunch, and the init servicing saying "no more".

100% success rate of the app loading as expected if I simply run a `snap restart onboard.onboard` or `systemctl retsart snap.onboard.onboard` after it fails and I ssh into the machine.

_Originally posted by @jrmcpeek in https://github.com/MirServer/ubuntu-frame-osk/issues/58#issuecomment-1730182354_
            

----

[edit]

The title may be misleading about the frequency of problems. In https://github.com/MirServer/ubuntu-frame/issues/151#issuecomment-1731894944 it says "one failed run across 30 power cycles"

The log included there shows that Frame segfaults, but doesn't contain anything to help identify why

AlanGriffiths:
We've been mostly getting away with drawing and flushing on random threads because these windows are rarely updated and we get consistency by dumb luck.

But if the timing is "just right" (see #151) all hell breaks loose.

Clearly there was a time when we thought about which calls need to be on the worker thread, because there was a semaphore to cause a flush, but that wasn't being used by the current code. (Logically, it should have been used in `FullscreenClient::draw()` as that doesn't run on the worker thread, but that doesn't actually fix #151 because `wl_surface_attach()` fails.)

Anyway, as we're not using `flush_signal`, and `draw_screen()` needs to be called on the worker thread, and the only place that isn't true is `draw()`, we can replace it with a `draw_signal` and do everything right.

Fixes: #151

### DIFF
--- a/src/background_client.cpp
+++ b/src/background_client.cpp
@@ -24,7 +24,6 @@
 #include <cstring>
 #include <codecvt>
 #include <sstream>
-#include <thread>
 
 #include <boost/throw_exception.hpp>
 #include <boost/filesystem.hpp>
--- a/src/egfullscreenclient.cpp
+++ b/src/egfullscreenclient.cpp
@@ -141,7 +141,7 @@ void egmde::FullscreenClient::Output::done(void* data, struct wl_output* /*wl_ou
 }
 
 egmde::FullscreenClient::FullscreenClient(wl_display* display, std::optional<Path> diagnostic_path, uint diagnostic_delay, miral::MirRunner* runner, WindowManagerObserver* window_manager_observer) :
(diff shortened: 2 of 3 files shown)


Saviq (approved):
I've not been able to reproduce here, but yeah this makes perfect sense.

PR #157, opened 2023-10-19
Add a --wallpaper-enabled argument to optionally disable wallpaper draws

mattkae:
## To test
See the wallpaper:
```
./frame
./frame --wallpaper-enabled=true 
```

Don't see the wallpaper:
```
./frame --wallpaper-enabled=false
```

## What's new?
- Added a `wallpaper_enabled` flag that is set to true by default
- Only doing things in `BackgroundClient::Self::draw_screen` if we have errors or the wallpaper is enabled
- If we have an error _and_ we don't want the wallpaper _but_ we don't have a diagnostic file, then we draw the wallpaper as we did before (this may be undesirable, but its a very rare case, and it maintains previous behavior)


Saviq (commented):
@mattkae were you able to confirm this DTRT as far as smooth boot is concerned?

If I leave plymouth running and `deactivate` it, then start Frame without the wallpaper on tty1, I get black immediately, until the timeout passes and we're back to the wallpaper. We want the splash screen to stay until we draw the client / wallpaper / diagnostic. Do we need to bring in some of #3033 for this?

That said, mir_demo_server behaves similarly - it goes to black straight up.

I'm on qemu, mind you, and see this warning, so maybe that's why?
```
 < -warning- > gbm-kms: Unable to determine the current display mode.
```

I'd _probably_ go for just `--wallpaper=<true|false>`, BTW.

AlanGriffiths:
> @mattkae were you able to confirm this DTRT as far as smooth boot is concerned?

I don't think this can be to address seamless boot directly: that needs the server to defer compositing until content from the client app is available. This just means that compositing will start without a background (as you say, just like mir_demo_server, or miriway without swaybg).

The idea might be to simplify detecting whether there's an app providing content by preventing the "wallpaper client" from providing content. But I'm not convinced that's necessary, or a good idea.

mattkae:
@Saviq @AlanGriffiths Alan is right on this one. The compositor itself is going to clear the screen once it is ready to start drawing surfaces. As Alan said, we could make it not do anything until it first has a "renderable" to render. Otherwise, we could do the deeper task of reading in the plymouth buffer and rendering that ourselves.

Saviq:
> I don't think this can be to address seamless boot directly: that needs the server to defer compositing until content from the client app is available. This just means that compositing will start without a background (as you say, just like mir_demo_server, or miriway without swaybg).

> As Alan said, we could make it not do anything until it first has a "renderable" to render.

Yeah that's what I think we should be doing. Off the top I can't think of why it couldn't be the default behaviour?

---

> The idea might be to simplify detecting whether there's an app providing content by preventing the "wallpaper client" from providing content. But I'm not convinced that's necessary, or a good idea.

Do you have another approach in mind? Why / how would we want to treat the background client in a special way? In truth, we could probably (when this, or similar, option is engaged) just delay the background client altogether by `$timeout`, no smaller than the diagnostic timeout?

---

> Otherwise, we could do the tougher task of reading in the plymouth buffer and rendering that ourselves until the user has supplied another renderable.

That wouldn't change much compared to the above, would it?

I mean, not until we start doing something more with the buffer.

Saviq:
OK so this, together with MirServer/mir#3086 looks exactly like what we need :)

I'm just trying to think about what's the best user experience. A `--wallpaper-timeout` came to mind, but then isn't the diagnostic what we'd like to see first, if configured? So the wallpaper timeout would have to be counted towards the diagnostic one.

But do we really need another timeout? The behaviour here, I think, is mostly correct - the first thing we see (with `--wallpaper=false`) is the diagnostic. We still see the wallpaper if there is no diagnostic text, so that's the icky bit - alternative would be to stay on the splash screen indefinitely, but maybe that's what we should do?

If someone disabled the wallpaper, and does not have the diagnostic set up - maybe the splash screen should stay on screen?

mattkae:
> OK so this, together with [MirServer/mir#3086](https://github.com/MirServer/mir/pull/3086) looks exactly like what we need :)
> 
> I'm just trying to think about what's the best user experience. A `--wallpaper-timeout` came to mind, but then isn't the diagnostic what we'd like to see first, if configured? So the wallpaper timeout would have to be counted towards the diagnostic one.
> 
> But do we really need another timeout? The behaviour here, I think, is mostly correct - the first thing we see (with `--wallpaper=false`) is the diagnostic. We still see the wallpaper if there is no diagnostic text, so that's the icky bit - alternative would be to stay on the splash screen indefinitely, but maybe that's what we should do?
> 
> If someone disabled the wallpaper, and does not have the diagnostic set up - maybe the splash screen should stay on screen?

Maybe if there isn't any diagnostic text but there _is_ some error, then we show some default diagnostic text? I agree that it feels weird to show the background out of nowhere when there's an error. I'm up for either some default text or just keeping the splash on screen. Although, leaving the splash on screen makes it a bit trickier to debug from a user's perspective

Saviq:
> Maybe if there isn't any diagnostic text but there _is_ some error, then we show some default diagnostic text? I agree that it feels weird to show the background out of nowhere when there's an error.

It's not that there is an error, necessarily. And even if there is - Frame can't know what it is.

> I'm up for either some default text or just keeping the splash on screen. Although, leaving the splash on screen makes it a bit trickier to debug from a user's perspective

Well, that's exactly what the diagnostic is for. The "user" in the case of Frame shouldn't be doing any debugging, just call the number that's on screen :)

Saviq:
And just to confirm we're DTRT:
The before:
[Nagranie ekranu z 2023-10-24 16-10-07.webm](https://github.com/MirServer/ubuntu-frame/assets/231077/76c26487-f3e4-44f2-93e1-f2f49caa4519)

Saviq:
And the after:
[Nagranie ekranu z 2023-10-24 16-02-07.webm](https://github.com/MirServer/ubuntu-frame/assets/231077/03e4e171-d90f-4a2b-a403-7f06d77e258a)

mattkae:
> Well, that's exactly what the diagnostic is for. The "user" in the case of Frame shouldn't be doing any debugging, just call the number that's on screen :)

That makes sense to me! So if they fail to set up the diagnostic _and_ the wallpaper is turned off, then we keep showing the splash. Does that sound right?

> And just to confirm we're DTRT:

And yup, that's what I would expect. The splash is shown all the way up to the first render of the app.

Saviq:
> That makes sense to me! So if they fail to set up the diagnostic _and_ the wallpaper is turned off, then we keep showing the splash. Does that sound right?

Yeah exactly.

> > And just to confirm we're DTRT:
> 
> And yup, that's what I would expect. The splash is shown all the way up to the first render of the app.

Oh I wasn't asking ;D

mattkae:
@Saviq I have implemented + tested only showing the diagnostic screen when we have diagnostic info :+1:

Saviq:
@mattkae that if/else dance looks like it could use some rework?

mattkae:
> @mattkae that if/else dance looks like it could use some rework?

As in its too verbose, or is it incorrect?

Saviq:
> As in its too verbose, or is it incorrect?

I mean this:

https://github.com/MirServer/ubuntu-frame/blob/9f247ce1b325f8b74fe3ddb7d400346a9e36232a/src/background_client.cpp#L401-L416

and then this:

https://github.com/MirServer/ubuntu-frame/blob/9f247ce1b325f8b74fe3ddb7d400346a9e36232a/src/background_client.cpp#L457-L467

The control flow got quite twisted?

Saviq (approved):
Looks good to me!

@AlanGriffiths would like your ACK in case I'm not thinking of something here…

src/background_client.cpp
```diff
@@ -105,6 +106,7 @@ struct BackgroundClient::Self : egmde::FullscreenClient
     
     void render_text(uint32_t width, uint32_t height, unsigned char* buffer) const;
 
+    bool wallpaper_enabled;
```
AlanGriffiths, line 109: This is never updated after initialization, so should be `const`.

src/background_client.cpp
```diff
@@ -377,6 +387,23 @@ void BackgroundClient::Self::draw_screen(SurfaceInfo& info, bool draws_crash) co
...
+    // Don't draw diagnostic background if file is empty or font not found
+    bool file_exists;
+    if (fs::exists(diagnostic_path.value_or("")))
+    {
+        file_exists = fs::file_size(diagnostic_path.value());
+    }
+    else
+    {
+        file_exists = false;
+    }
```
AlanGriffiths, line 399:
```suggestion
    bool const file_exists = diagnostic_path && fs::exists(diagnostic_path.value()) && fs::file_size(diagnostic_path.value());
```

AlanGriffiths: Although, maybe the variable should be `have_diagnostic` - the "file" is an implementation detail.

src/background_client.cpp
```diff
@@ -398,21 +398,10 @@ void BackgroundClient::Self::draw_screen(SurfaceInfo& info, bool draws_crash) co
         file_exists = false;
     }
 
-    if (!wallpaper_enabled)
+    bool should_draw_crash = draws_crash && file_exists;
```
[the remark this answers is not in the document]
AlanGriffiths: Maybe the variable should be `should_show_diagnostic`?

AlanGriffiths (changes requested):
Some small `const` cleanup and some (non blocking) bikeshedding about the naming

AlanGriffiths (approved):
LGTM

PR #158, opened 2023-11-02
snap: enable smooth boot through system-files

Saviq:


Saviq:
Asked for autoconnection here:

https://forum.snapcraft.io/t/autoconnect-for-ubuntu-frame-system-files/37582

PR #164, opened 2024-01-19
Drop some deprecated and/or obsolete stuff (and fix edge)

AlanGriffiths:
Also fixed the `edge` builds, by staging `libmiral{6,7}` indirectly. (The old trick of `try: [libmiral7]/else: [libmiral6]` is no longer supported by snapcraft


Saviq (approved):
Sures.

Issue #173, opened 2024-03-18
Add way to change display configuration without snapd


CharleeSF:
We are trying to create an image that does all configuration on first startup.
The device should not require CLI interaction for everything to be setup correctly.
Since the device is in portrait mode we need to change the ubuntu-frame configuration.
However, this is only possible from command line (or through the snapd interface).

Please provide a way to interact with ubuntu-frame configuration through an interface.
For example, you can provide a content interface in which ubuntu-frame drops its latest display config, and where it can pick up new display configurations.

I tried to look if something like this is already available, and it seems like something similar to what I want is done with ```display_file="$SNAP_DATA/frame.display"```.
However, I cannot touch that file from another snap.

Saviq:
@CharleeSF since you're building an image, you can use the default config on the gadget?

https://mir-server.io/docs/configuring-ubuntu-frame-through-a-gadget-snap

We'd rather not expose the full display config to be changed runtime, as this is meant to be static - but we're planning to expose the layout option, likely through a content interface.

CharleeSF:
Ok, thanks!

The example helps :) The multi-line config option can be a bit of a struggle!
Also didn't know you could specify only part of the file. I'm gonna try this.

CharleeSF:
Dear @Saviq 

Just wondering, the example you gave has the value "unknown-1" for the display.
I was hoping that somehow ubuntu-frame would fill this in by found values, but it seems this is not the case.
Rather, it seems to expect the value of the display:

Does this mean that I need to make a separate gadget snap for every hardware I have? What if I want to test my snap on Qemu? The display variables are different?

If the 'auto-infer' logic isn't working when setting default values, I think there is still value in having a way to read and write ubuntu-frame's config from another snap.

Saviq:
@CharleeSF the output names are stable, and something like `HDMI-1`, `DisplayPort-2`. This is most often enough to keep a stable display config.

You can have entries for outputs not present on a piece of hardware, so if you have different hardware (say, one has HDMI, another DisplayPort, yet another an unused port so you need `HDMI-2`). You can have a single image populated with them all (possibly using YAML anchors to avoid duplication) and only the present, active ones will be used.

We did consider, just didn't get to, yet, matching the displays by something else than output name (model, serial number etc.). If you can describe your use case, we will take that into account.

Another option we could think of is allowing wildcards, so you could say something like:

```
# ...
- card-id: 0
  *:
    orientation: left
```

[...]

PR #174, opened 2024-03-18
Add content interface and file-based display configuration

CharleeSF:
Hey there,

This PR adds a way to configure the ubuntu-frame display variable from another snap without the snapd-control interface.

Any suggestions to improve are welcome.

Will solve issue #173


ogra1:
You might want to drop the name change from your snapcraft.yaml to actually have a chance to get it merged ;)

PR #185, opened 2024-05-28
Tidy up after static analysis

AlanGriffiths:
Fix true positives

### DIFF
--- a/src/background_client.cpp
+++ b/src/background_client.cpp
@@ -269,7 +269,7 @@ void BackgroundClient::render_background(
 {
     Colour new_pixel;
 
-    for (auto current_y = 0; current_y < height; current_y++)
+    for (uint32_t current_y = 0; current_y < height; current_y++)
     {
         // Render gradient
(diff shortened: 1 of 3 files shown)


mattkae (approved):
The code looks good and makes sense :+1: I am unsure how to test `check_for_exposed_outputs` properly though. I can trust that it works, but if you have a simple repo that I could try that would be good too

AlanGriffiths:
> The code looks good and makes sense 👍 I am unsure how to test `check_for_exposed_outputs` properly though. I can trust that it works, but if you have a simple repo that I could try that would be good too

As previously discussed, we need better test rigging in this area. And, I think, the logic for overlapping, non-identical, outputs is suspect.

I trust it is clear that this rework doesn't break anything that worked before. (I could drop [59868aa](https://github.com/canonical/ubuntu-frame/pull/185/commits/59868aa8061894376ddf1aaf52b263ae93d03751), the "extract method" refactor should be uncontentious.

PR #188, opened 2024-06-17
display: run configure script after populating the config

### LINKED ISSUE #187: display: populates `display` config value with stale configuration
> 1) I had monitor with preferred video mode "1280x800@52.9" connected then  I shut down the board.
> 2) Connected different monitor with preferred mode "1366x768@59.94Hz" and rebooted the board.
> 3) Then I executed above steps but "display" still shows the same old config.
> 4) Also snap logs ubuntu-frame did show that ubuntu-frame has detected new monitor with preferred timing of  "1366x768@59.94Hz"

Looks like we're using the old config YAML in this case to populate the `display` value.

Saviq:
Otherwise the file doesn't get the `DO NOT EDIT` header and things don't work as expected.

Fixes #187


AlanGriffiths:
Does this actually do the best thing?

The dance between `run-daemon`, `configure` and `initialise-display-config` already looks fragile. And executing `configure` to get the side effect of rewriting `$SNAP_DATA/frame.display` seems to add complexity to that.

It looks like the `elif grep "DO NOT EDIT THIS FILE BY HAND"...` clause in the `configure` script is the better place to handle this case. (By removing the file regardless of the header.)

Saviq:
> Does this actually do the best thing?
> 
> The dance between `run-daemon`, `configure` and `initialise-display-config` already looks fragile. And executing `configure` to get the side effect of rewriting `$SNAP_DATA/frame.display` seems to add complexity to that.

On the one hand, yes - on the other, `configure` should be idempotent, and safe… it already ran on refresh / installation just before this script. If we're not confident in it, that needs fixing.

I could just rewrite the file directly in the `initialise-` script, but that risks them going out of sync…

> It looks like the `elif grep "DO NOT EDIT THIS FILE BY HAND"...` clause in the `configure` script is the better place to handle this case. (By removing the file regardless of the header.)

That would cause custom configuration (predating the `display` key) to be overwritten on the first run of the `configure` script e.g. on refresh. And we need to backport the fix to the `22` track.

AlanGriffiths (approved):
OK

PR #190, opened 2024-06-17
display: run configure script after populating the config (#188)

### LINKED ISSUE #187: display: populates `display` config value with stale configuration

Saviq:
Otherwise the file doesn't get the `DO NOT EDIT` header and things don't work as expected.

Fixes #187


Saviq:
@AlanGriffiths go? No go?

This already got merged in `24/` and `22/`.

AlanGriffiths (approved):

PR #205, opened 2024-10-11
Adjusting code to account for the fact that the returned app_id no longer ends with the .desktop suffix

mattkae:
relies on https://github.com/canonical/mir/pull/3629

## What's new?
- We dont' want a `.desktop` file suffix in the applicaiton id

### DIFF
--- a/launcher/lib/services/desktop_file_manager.dart
+++ b/launcher/lib/services/desktop_file_manager.dart
@@ -103,12 +103,16 @@ class DesktopFileManager {
       return null;
     }
 
-    if (!f.path.endsWith('.desktop')) {
+    const String desktopFileSuffix = '.desktop';
+    if (!f.path.endsWith(desktopFileSuffix)) {
       return null;
     }
 
-    final desktopFile = ApplicationFileDesktopFile(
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
> relies on [canonical/mir#3629](https://github.com/canonical/mir/pull/3629)

We probably want to land code that works both with and without canonical/mir#3629 - otherwise we'd be be blocked from updating Frame without first releasing Mir

Saviq:
@mattkae ping!

We just need to _strip_ `.desktop` from the app id here, no?

mattkae:
> @mattkae ping!
> 
> We just need to _strip_ `.desktop` from the app id here, no?

Yup! That's what I did in this PR

AlanGriffiths:
Maybe I misunderstand, doesn't this still _require_ the ".desktop" extension?

mattkae:
> Maybe I misunderstand, doesn't this still _require_ the ".desktop" extension?

On the frontend, we read all of the `X.desktop` files. The app id is `X` in this case, unless otherwise specified inside of the desktop file. We then resolve `X` instead of `X.desktop`

AlanGriffiths (approved):
OK, got it

PR #215, opened 2025-02-26
Set XDG_CURRENT_DESKTOP

kenvandine:
Set XDG_CURRENT_DESKTOP to UbuntuFrame:Mir to allow showing and hiding desktop files based on session.


AlanGriffiths (approved):
Looks sensible.

My only concern is that starting to check this might hide things we show now. But that is not yet a supported scenario

PR #218, opened 2025-04-01
Implement custom client placement

mattkae:
Relies on https://github.com/canonical/mir/pull/3869

## What's new?
- Implemented `DisplayConfiguration::set_layout_userdata_builder` such that it builds a `LayoutMetadata`, which parses the layout display configuration into a list of application placement data
- When placing a new surface, we try and place it according to `LayoutMetadata` first. If that fails, we try to assign it to an output if it is mapped to one. Otherwise we just fullscreen it as we did before.
- Warning the user when their surface placement does not intersect any output
- Warning the user when they have specified that their surface should be placed both in a particular position and on a particular card

## To test
1. Have a `~/.config/frame.display` that looks like:
    ```yaml
    layouts:
    # keys here are layout labels (used for atomically switching between them).
    # The yaml anchor 'the_default' is used to alias the 'default' label
    
      default:
        cards:
        # a list of cards (currently matched by card-id)

        - card-id: 0
          unknown-1:
            # This output supports the following modes: 1280x1024@60.0
            #
            # Uncomment the following to enforce the selected configuration.
            # Or amend as desired.
            #
            state: enabled	# {enabled, disabled}, defaults to enabled
            mode: 1280x1024@60.0	# Defaults to preferred mode
            position: [0, 0]	# Defaults to [0, 0]
            orientation: normal	# {normal, left, right, inverted}, defaults to normal
            scale: 1
            group: 0	# Outputs with the same non-zero value are treated as a single display

        applications:
          # Two applications side by side on the first display.
          - snap-name: gedit
            position: [0, 0]
            size: [780, 1024]
          - surface-title: Mines
            position: [780, 0]
            size: [500, 1024]
    ```
2. Run frame with:
    ```
    frame --display-layout=default
    ```
3. Run `gedit` and `gnome-mines`
4. See that they are positioned correctly
5. Run another app and see that it takes up the full screen

Screenshot from 2025-04-02 14-58-04

## TODO Later
- For now, we are ignoring the requirement around the Z-order of applications

### DIFF
--- a/src/frame_window_manager.cpp
+++ b/src/frame_window_manager.cpp
@@ -181,13 +181,12 @@ void FrameWindowManagerPolicy::handle_layout(
 
     auto const snap_instance_name = snap_instance_name_of(application);
     auto const surface_title = specification.name() ? specification.name() : window_info.name();
-    std::shared_ptr<LayoutMetadata> layout_metadata;
 
 #if MIRAL_MAJOR_VERSION > 5 || (MIRAL_MAJOR_VERSION == 5 && MIRAL_MINOR_VERSION >= 3)
+    std::shared_ptr<LayoutMetadata> layout_metadata;
     auto const layout_userdata = display_config.layout_userdata("applications");
     if (layout_userdata.has_value())
         layout_metadata = std::any_cast<std::shared_ptr<LayoutMetadata>>(layout_userdata.value());
-#endif
 
     // If the snap name or surface title is mapped to a particular position and size, then the surface is placed there.
     if (layout_metadata && layout_metadata->try_layout(specification, surface_title, snap_instance_name))
@@ -216,6 +215,7 @@ void FrameWindowManagerPolicy::handle_layout(
 
         return;
     }
+#endif
 
     // If the snap name or surface title is mapped to a particular output, then the surface is fullscreen on that output.
     if (assign_to_output(specification, surface_title, snap_instance_name))
--- a/src/frame_window_manager.cpp
+++ b/src/frame_window_manager.cpp
@@ -185,6 +185,7 @@ void FrameWindowManagerPolicy::handle_layout(
     // If the snap name or surface title is mapped to a particular position and size, then the surface is placed there.
     if (layout_metadata != nullptr && layout_metadata->try_layout(specification, surface_title, snap_instance_name))
     {
+        // Let's warn if the user is placing their surface beyond the extents of all outputs
         Rectangle const extents(specification.top_left().value(), specification.size().value());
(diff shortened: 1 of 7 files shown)


src/layout_metadata.cpp
```diff
@@ -0,0 +1,164 @@
...
+    });
+}
+
+bool LayoutMetadata::try_layout(miral::WindowSpecification& specification,
+    mir::optional_value<std::string> const& title,
+    std::string_view snap_name) const
+{
+    for (auto const& app : applications)
+    {
+        if (app.snap_name == snap_name || app.surface_title == title)
```
robert-ancell, line 77: If both `app.snap_name` or `snap_name` are unset/blank, will this hit a false positive?

src/frame_window_manager.cpp
```diff
@@ -162,36 +169,85 @@ bool FrameWindowManagerPolicy::handle_keyboard_event(MirKeyboardEvent const* eve
...
-        WindowInfo window_info{};
-        if (needs_bespoke_fullscreen_placement(specification, window_info))
+        // Let's warn if the user is placing their surface beyond the extents of all outputs
+        Rectangle const extents(specification.top_left().value(), specification.size().value());
+        bool found = false;
+        for (auto const& output : active_outputs)
         {
-            auto const snap_instance_name = snap_instance_name_of(app_info.application());
+            if (output.extents().overlaps(extents))
+                found = true;;
```
robert-ancell, line 197: Nit: duplicate `;;`

src/frame_window_manager.cpp
```diff
@@ -59,18 +59,21 @@ bool needs_bespoke_fullscreen_placement(WindowSpecification& spec, WindowInfo co
...
-    case mir_window_state_minimized:
-    case mir_window_state_hidden:
-    case mir_window_state_attached:return false;
-
-    default:;
+        case mir_window_state_minimized:
+        case mir_window_state_hidden:
+        case mir_window_state_attached:
+            return false;
+        default:;
```
robert-ancell, line 66: Nit: Semicolon on end (from existing code).

robert-ancell (approved):
Looks good!

src/layout_metadata.cpp
mattkae: I think this is only true for `title`, since that can be optional. `snap_name` is at minimum an empty string, and if someone actually wanted to match on an empty string, that isn't invalid. However, that could never happen because we would tell them that it's an error when we read it from the YAML originally :)

The case where `title` is unset is accurate though and I will update

PR #219, opened 2025-04-10
Enabling unit tests using gtest in the project

mattkae:


src/CMakeLists.txt
```diff
@@ -4,33 +4,54 @@ include(FindPkgConfig)
...
-add_executable(frame
-    frame_main.cpp
+add_library(frame-implementation
     frame_authorization.cpp frame_authorization.h
     frame_window_manager.cpp frame_window_manager.h
     egfullscreenclient.cpp egfullscreenclient.h
     background_client.cpp background_client.h
-    snap_name_of.cpp snap_name_of.h
-)
+    snap_name_of.cpp snap_name_of.h)
```
robert-ancell, line 19: Nit: Did you intentionally change the closing parenthesis from being on a separate line? The existing convention seems to be for multi-line rules to have it as it was originally.

robert-ancell (approved):
My CMake skills are pretty limited, but the change makes sense to me and it will be handy to be able to add tests like this.

PR #222, opened 2025-04-28
(#221) windows are now clipped to their tile if they are absolutely positioned and sized

### LINKED ISSUE #221: Apps going fullscreen despite custom size
Steps:
- configure Frame to place mir-kiosk-neverputt custom
  https://mir-server.io/docs/how-to-configure-ubuntu-frame-for-multiple-outputs#heading--custom-client-placement
   ```yaml
   layouts:
     default:
       cards:
       - card-id: 0

       applications:
       - snap-name: mir-kiosk-neverputt
         position: [640, 0]
         size: [640, 512]
   ```
- start frame and neverputt

Expected:
- placed at the configured layout

Current:
- decorations shown at the current layout (#220)
- app shown in fullscreen size, positioned according to layout

Image

mattkae:
fixes #221 

## Problem
- Some clients will not abide by our wishes for them to be a particular size

## Solution
- Even if the buffer is too large, we set the clip area accordingly so that the client at least does not overflow outside of the tile

### DIFF
--- a/src/frame_window_manager.cpp
+++ b/src/frame_window_manager.cpp
@@ -183,13 +183,8 @@ void FrameWindowManagerPolicy::handle_layout(
     auto const surface_title = specification.name() ? specification.name() : window_info.name();
 
 #if MIRAL_MAJOR_VERSION > 5 || (MIRAL_MAJOR_VERSION == 5 && MIRAL_MINOR_VERSION >= 3)
-    std::shared_ptr<LayoutMetadata> layout_metadata;
-    auto const layout_userdata = display_config.layout_userdata("applications");
-    if (layout_userdata.has_value())
-        layout_metadata = std::any_cast<std::shared_ptr<LayoutMetadata>>(layout_userdata.value());
-
     // If the snap name or surface title is mapped to a particular position and size, then the surface is placed there.
-    if (layout_metadata && layout_metadata->try_layout(specification, surface_title, snap_instance_name))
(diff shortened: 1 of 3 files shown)


src/frame_window_manager.cpp
```diff
@@ -264,6 +266,34 @@ auto FrameWindowManagerPolicy::place_new_window(ApplicationInfo const& app_info,
...
+#if MIRAL_MAJOR_VERSION > 5 || (MIRAL_MAJOR_VERSION == 5 && MIRAL_MINOR_VERSION >= 3)
+    // After a window has been placed at a specific coordinate, we must clip it to its tile
+    // so that it does not overlap with other applications in the event that the client insists on
+    // submitting buffers that are larger than its tile.
+    auto const application = window_info.window().application();
+    auto const snap_instance_name = snap_instance_name_of(application);
+    auto const surface_title = window_info.name();
+
+    std::shared_ptr<LayoutMetadata> layout_metadata;
+    auto const layout_userdata = display_config.layout_userdata("applications");
```
robert-ancell, line 280: This is a duplicated in handle_layout - suggest moving into a `get_layout()` method. It's also unclear why "applications" means it has a layout - using a method or comment might make this clearer. Is it possible that layout information might be provided in the future by another method that doesn't use the "applications" data?

robert-ancell (approved):
Changes make sense to me, just a small improvement that I've suggested.

PR #223, opened 2025-05-02
(#220): applications placed exactly are now placed as fullscreen windows, but do not get placed and sized according to the fullscreen paradigm

### LINKED ISSUE #220: `applications:` need to think they're fullscreen still
Steps:
- configure Frame to place a snap/app custom
  https://mir-server.io/docs/how-to-configure-ubuntu-frame-for-multiple-outputs#heading--custom-client-placement
- start the configured app

Expected:
- no window decorations

Current:
- decorations (e.g. scummv)

screenshot

mattkae:
fixes #220 

## What's new?
- Refactored `try_position_exactly` so that the logic is more confined
- During `confirm_placement_on_display` and `advise_end`, we now check if we are positioned exactly before trying to make the window take up the full width and height

### DIFF
--- a/src/frame_window_manager.cpp
+++ b/src/frame_window_manager.cpp
@@ -179,12 +179,9 @@ void FrameWindowManagerPolicy::handle_layout(
     if (!can_position_be_overridden(specification, window_info))
         return;
 
-    auto const snap_instance_name = snap_instance_name_of(application);
+    auto const snap_instance_name = application ? snap_instance_name_of(application) : "";
     auto const surface_title = specification.name() ? specification.name() : window_info.name();
-
-#if MIRAL_MAJOR_VERSION > 5 || (MIRAL_MAJOR_VERSION == 5 && MIRAL_MINOR_VERSION >= 3)
-    // If the snap name or surface title is mapped to a particular position and size, then the surface is placed there.
-    if (try_position_exactly(specification, snap_instance_name, surface_title.value_or("")))
+    if (try_position_exactly(specification, window_info, application))
     {
         // Let's warn if the user is placing their surface beyond the extents of all outputs
         Rectangle const extents(specification.top_left().value(), specification.size().value());
@@ -212,7 +209,6 @@ void FrameWindowManagerPolicy::handle_layout(
             window_info.clip_area(extents);
         return;
     }
-#endif
 
     // If the snap name or surface title is mapped to a particular output, then the surface is fullscreen on that output.
     if (assign_to_output(specification, surface_title, snap_instance_name))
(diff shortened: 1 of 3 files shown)


Saviq:
@mattkae you might want to bribe someone to review this ;)

But, could we add some tests while we're here?

src/frame_window_manager.cpp
```diff
@@ -179,12 +179,9 @@ void FrameWindowManagerPolicy::handle_layout(
     if (!can_position_be_overridden(specification, window_info))
         return;
 
-    auto const snap_instance_name = snap_instance_name_of(application);
+    auto const snap_instance_name = application ? snap_instance_name_of(application) : "";
```
robert-ancell, line 182: I don't think application can be nullptr, as it's passed as a reference.

src/frame_window_manager.cpp
```diff
@@ -519,17 +515,25 @@ bool FrameWindowManagerPolicy::PlacementMapping::set_output_for_snap(
...
 bool FrameWindowManagerPolicy::try_position_exactly(
     WindowSpecification& spec,
-    std::string const& snap_instance_name,
-    std::string const& surface_title) const
+    WindowInfo const& window_info,
+    Application const& application) const
 {
+#if MIRAL_MAJOR_VERSION > 5 || (MIRAL_MAJOR_VERSION == 5 && MIRAL_MINOR_VERSION >= 3)
     /// Retrieve the layout information from the "applications" key in the layout's userdata.
+    auto const snap_instance_name = application ? snap_instance_name_of(application) : "";
```
robert-ancell, line 523: applications can't be nullptr, right?

robert-ancell (approved):
Looks cleaner. I think some of the nullptr checks aren't required though? Certainly a test would be nice :)

src/frame_window_manager.cpp
mattkae: Application is actually just a rename of shared_ptr :)

```c++
using Application = std::shared_ptr<mir::scene::Session>;
```

mattkae:
> @mattkae you might want to bribe someone to review this ;)
> 
> But, could we add some tests while we're here?

I will actually make a ticket for testing so that we can do it right. Ideally, I would use the new WindowManagementTestHarness to facilitate testing here, but that is waiting on packaging the stub platforms. Once that's published, we should use those tests and try and test all of frame :)

mattkae:
> @mattkae "This branch has conflicts that must be resolved"

Fixed!

AlanGriffiths (changes requested):
There's a merge conflict and some inconsistency in the minimal Mir version supported by `#if`s

Also the MR title is hard to parse. Is it "Applications placed exactly are in the fullscreen state"?

src/frame_window_manager.cpp
```diff
@@ -519,13 +515,21 @@ bool FrameWindowManagerPolicy::PlacementMapping::set_output_for_snap(
 
 bool FrameWindowManagerPolicy::try_position_exactly(
     WindowSpecification& spec,
-    std::string const& snap_instance_name,
-    std::string const& surface_title) const
+    WindowInfo const& window_info,
+    Application const& application) const
 {
+#if MIRAL_MAJOR_VERSION > 5 || (MIRAL_MAJOR_VERSION == 5 && MIRAL_MINOR_VERSION >= 3)
```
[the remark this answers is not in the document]
mattkae: I moved them down into the method instead of having them scattered everywhere. The version check only needs to happen in a single place :)

src/frame_window_manager.cpp
```diff
@@ -519,17 +516,25 @@ bool FrameWindowManagerPolicy::PlacementMapping::set_output_for_snap(
 
 bool FrameWindowManagerPolicy::try_position_exactly(
     WindowSpecification& spec,
-    std::string const& snap_instance_name,
-    std::string const& surface_title) const
+    WindowInfo const& window_info,
+    Application const& application) const
 {
+#if MIRAL_VERSION >= MIR_VERSION_NUMBER(5, 3, 0)
```
AlanGriffiths, line 522: I think we can drop this check. The current Mir release has Miral 5.5, and that's what Frame is supported against

AlanGriffiths (approved):
Okay, but we could cleanup some legacy conditionals too

PR #227, opened 2025-05-28
screenshot: error out when `wayplug` is not connected

### LINKED ISSUE #224: `ubuntu-frame.screenshot` doesn't work on classic
Per title, try `ubuntu-frame.screenshot` to no result:

```
$ WAYLAND_DISPLAY=wayland-1 ubuntu-frame.screenshot
failed to create display
```

Doesn't seem to be auth problems, as `ubuntu-frame-osk` works fine.

Saviq:
Fixes #224

### DIFF
--- a/scripts/bin/screenshot
+++ b/scripts/bin/screenshot
@@ -1,6 +1,16 @@
 #!/bin/sh
 set -eu
 
+if ! snapctl is-connected wayplug; then
+    echo "ERROR: The wayland plug needs to be connected:" >&2
+    echo "  # on Classic" >&2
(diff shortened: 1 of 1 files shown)


Saviq:
An alternative could be for `launcher` and `screenshot` to `slots: [wayland]`, then any connection to the `ubuntu-frame:wayland` slot would also fulfil the requirement.

AlanGriffiths:
> An alternative could be for `launcher` and `screenshot` to `slots: [wayland]`, then any connection to the `ubuntu-frame:wayland` slot would also fulfil the requirement.

The best answer is to fix the `wayland` interface so that it autoconnects with multiple providers.

PR #233, opened 2025-06-10
docs: use intersphinx links across the board and improve `conf.py`

Saviq:


mattkae (approved):
I don't think that I (or you) missed anything, looks good to me :)

PR #235, opened 2025-07-16
doc: fix notes, lexing and add a web kiosk support note

Saviq:


docs/how-to/packaging-iot-gui/a-flutter-application.md
```diff
@@ -40,16 +40,15 @@ Open a terminal window and type:
 sudo snap install ubuntu-frame --channel=24
 ```
 
-[note status="channel=24"]
+```{admonition} channel=24
```
keirthana, line 43: Do you want to use admonitions with custom titles? Just trying to understand the reasons behind multiple different types of admonition.

For example, there is a note higher up in this page where it just says "Note" in the text format.

If you want a generic reference to follow across the documentation, I recommend using the format as demonstrated in https://github.com/canonical/sphinx-docs-starter-pack/blob/main/docs/reference/doc-cheat-sheet.rst#notes (see the raw format for exact syntax)

keirthana: It is not a deal-breaker. Custom titles attract attention from specific audience (those who care about channel=24, others will skip over it). In cases where we are writing information for specific audience, we can keep the custom titles.

In this particular case, it seems like a more generic note about why we are using channel 24. So we can drop it IMO. 

To be clear, this is not a deal-breaker and does not require a doc-wide update immediately but something to fix when you spot it.

Saviq: Replaced with standard notes :+1:

PR #247, opened 2025-09-02
Rework and rename the developer guide

keirthana:
The developer guide is a really good tutorial for developers looking to deploy their first graphical application with Ubuntu Frame. This PR tightens it a bit by removing redundant or unnecessary content (we can assume a basic level of technical knowledge from our users given that they are developers). 

The guide is also renamed appropriately.


docs/tutorial/deploy-graphical-application-with-frame.md
```diff
@@ -89,51 +71,29 @@ Now Frame's window should contain the "Mastermind" game.
...
-Close that (`Ctrl-Q`) and try the next example.
-
-#### SDL2 example: Neverputt
-
-```
-sudo apt install neverputt
-frame-it neverputt
-```
-
-Now Frame's window should contain the "Neverputt" game. You’ll see is that the game doesn’t fill the display. That’s because the application doesn’t understand Ubuntu Frame telling it to fill the screen. This is a problem with some applications and, in this case, can be fixed by editing `~/.neverball/neverballrc` [sic] to say “fullscreen 1” and restarting the game. (The same file works for neverball, but you may need to figure out the right configuration option for your application.) This is what you will see:
```
keirthana, line 114: Repeating the same instructions for different applications word for word is not necessary. Since the commands are exactly the same except for the name of the application, we can allow the user to experience the learning by try out the other two applications themselves.

I have still retained the parts that allow the user to troubleshoot and experiment.

PR #252, opened 2025-09-03
Replace architecture diagram with Mermaid diagram

keirthana:
Replace the kiosk architecture diagram with one created in Mermaid. The source is also added in the `.mmd` file.


keirthana:
> The point of the diagram is to show the different tech stacks:
> 
> 1. Toolkit with wayland support
> 2. Toolkit with only X11 support
> 3. Not sure what this is intended to be: Old version of electron?
> 
> By shrinking all three cases to a blank box this distinction is lost
> 
> (I do think we need to check on the intent of 3 though - I think it is now obsolete)

@AlanGriffiths I don't get your point about a blank box. Is this not how you see the diagram?

image

There are two boxes - one showing the case for toolkit with Wayland support and the other two cases inside another box. If the 3 one is not needed, we can remove that but I still think this image differentiates the two different tech stacks - with Wayland support and without. The only parts that are collapsed are that the application and the toolkit boxes - is that what you want to be differentiated?

Saviq:
> Is this not how you see the diagram?

That's probably the issue, in (my) Firefox it looks like this:
image

AlanGriffiths:
> If we end up with an image, I don't think there's a point in using Mermaid for it?

Agreed, Mermaid is very limiting: e.g. we can't decorate the diagram with lines (like the "snapd wayland interface") in the original diagram

keirthana:
> > If we end up with an image, I don't think there's a point in using Mermaid for it?
> 
> Agreed, Mermaid is very limiting: e.g. we can't decorate the diagram with lines (like the "snapd wayland interface") in the original diagram

If Mermaid doesn't work, we don't have to stay tied to Mermaid. If we are okay with using the end result image, I can use Google slides or something similar to recreate the diagram quickly for now.

AlanGriffiths:
> Let me know what you think.

The line placement is a bit unbalanced, but the content is fine. Do we need to introduce another place to manage our docs (assets manager)?

AlanGriffiths:
Also the title and description are obsolete - they both mention mermaid

keirthana:
> Also the title and description are obsolete - they both mention mermaid

Updated the title and description.

Saviq (changes requested):
> Yes, it works. You can check in the preview, if you'd like.

Right, with a white background it works :)

I suppose I was hoping for a dark background in dark mode :shrug:

Can you please install `git-lfs` and commit it from scratch? We're using Git Large File Storage for these.

Edit:
Might want to `git lfs track *.odp`, too.

AlanGriffiths (approved):
The diagram still looks asymmetrical - the vertical line is not centered. But losing case 3 is a win.

Not sure dark mode is improved though

PR #255, opened 2025-09-03
Adding release notes to ubuntu frame

mattkae:


docs/release-notes.md
```diff
@@ -0,0 +1,10 @@
+(release-notes)=
+
+# Release notes
+This pages provides release notes for Ubuntu Frame. Ubuntu frame is released
```
keirthana, line 4: Just a nit, we can drop the first statement as it is stating the obvious.

The page will get filled moving forward and won't look as sparse.

keirthana (approved):
Approving with a nitpick- up to you to take it or leave it.

PR #256, opened 2025-09-10
A merged how-to guide

keirthana:
This PR has a first look of how the following how-to guides can be merged:

- Packaging a Flutter application
- Packaging a GTK3 application
- Packaging an SDL2 application
- Packaging a Qt5 application
- Packaging an Electron application
- Packaging an X-11 application
We don't have to demo every single example in a how-to guide, we can trust the users to figure that out.

<s>To be resolved:

There is a pending question about whether the "Checking your application works with Ubuntu Frame" applies for Flutter and X-11 applications. </s>

This PR also does not merge the migrating a windows application how-to guide as there are more differences to that how-to guide although it doesn't deviate significantly in structure.

If this proposal is good, I will remove the individual how-to guides and add redirects for the deleted stuff.

### DIFF
--- a/docs/how-to/packaging-iot-gui/packaging-an-application.md
+++ b/docs/how-to/packaging-iot-gui/packaging-an-application.md
@@ -65,10 +65,6 @@ If you don't have git installed, now is a good time to install it (on Ubuntu, us
 
 ## Checking your application works with Ubuntu Frame
 
-```{important}
-If you are packaging an X11 based application, it won't yet work with Ubuntu Frame. You can skip this section and start packaging your application as a snap.
-```
-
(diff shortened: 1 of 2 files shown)


Saviq:
> We don't have to demo every single example

Still, it might be nice to mix it up? Use different examples for each step that has a demo?

keirthana:
> > We don't have to demo every single example
> 
> Still, it might be nice to mix it up? Use different examples for each step that has a demo?

We can, I was worried it could be confusing

AlanGriffiths:
> Something about it broke the doc build - will have a look next week (if nobody gets there first)

It wasn't me! (Not sure of the cause nor fix)

@keirthana can you pull in my changes (and adapt as you feel necessary)?

Saviq:
> What do you mean it is a GitHub issue?

This worked until yesterday, but suddenly started failing trying to fetch the LFS objects. I filed a support request with GitHub, have added you to CC now, not sure if you can see it at https://support.github.com/tickets/personal/0, but you should get notified about any activity from now on.

AlanGriffiths (approved):
I think this is good to go (but it would be better if the docs still built)

PR #264, opened 2025-09-17
(#262) only relayout a window if the client requests for its state, position, or size to change

mattkae:
fixes canonical/mir#4282 

## What's new?
We were applying the layout to a the window on every surface modification, which isn't right. Instead, it makes sense to apply it whenever the client attempts to change the state, size, or position of the surface. That way, we force the surface into the right position again.

### DIFF
--- a/src/frame_window_manager.cpp
+++ b/src/frame_window_manager.cpp
@@ -297,6 +297,7 @@ void FrameWindowManagerPolicy::handle_modify_window(WindowInfo& window_info, Win
 {
     WindowSpecification specification = modifications;
 
+    // FIXME: this shouldn't be necessary, see canonical/mir#4282
     // If the client requests a change to its state, size, or topleft, we for a
     // relayout so that it is aware of its true parameters.
     if (specification.state().is_set() || specification.size().is_set() || specification.top_left().is_set())


mattkae:
> > on every surface modification
> 
> Hmm, but why would `::handle_modify_window` be invoked for every frame? Is it a client problem?
> 
> I mean, should we be diffing the window specification in Mir _before_ calling `::handle` for it?
> 
> At least for Neverputt, the only communication going through is:
> 
> ```
> wl_buffer@27.release()
> wl_display@1.delete_id(29)
> wl_callback@29.done(26559134)
>  -> wl_surface@22.frame(new id wl_callback@29)
>  -> wl_surface@22.attach(wl_buffer@32, 0, 0)
>  -> wl_surface@22.damage(0, 0, 2147483647, 2147483647)
>  -> wl_surface@22.commit()
> wl_buffer@28.release()
> wl_display@1.delete_id(29)
> wl_callback@29.done(26559151)
>  -> wl_surface@22.frame(new id wl_callback@29)
>  -> wl_surface@22.attach(wl_buffer@27, 0, 0)
>  -> wl_surface@22.damage(0, 0, 2147483647, 2147483647)
>  -> wl_surface@22.commit()
> ```
> 
> I see no reason to be invoking `handle_modify_window`?

You are right. I should have dug deeper into the `modifications`. We are just calling this method in response to any `commit`, which seems a bit unreasonable

Saviq:
> @Saviq however, I _do_ think that my check still stands. If the window changes something else (e.g. userdata), then we should not care to relayout the surface

Well, yes, but then we'll add another member to `WindowSpecification` that should also cause a re-layout, and we'll forget to add another condition here ;)

I would be happier with "if only these properties that I know I'm not using in the re-layout are set" - but that just has the reverse of the same problem…

NB: I'm not saying we shouldn't get this in - it's a workaround for an existing Mir issue - but I'm not sure it should be the default.

src/frame_window_manager.cpp
```diff
@@ -296,7 +296,14 @@ void FrameWindowManagerPolicy::advise_delete_window(WindowInfo const& window_inf
 void FrameWindowManagerPolicy::handle_modify_window(WindowInfo& window_info, WindowSpecification const& modifications)
 {
     WindowSpecification specification = modifications;
-    handle_layout(specification, window_info.window().application(), window_info);
+
```
Saviq, line 299:
```suggestion

// FIXME: this shouldn't be necessary, see canonical/mir#4282
```

Saviq (approved):
Let's workaround!

PR #295, opened 2026-03-24
[safe mode] show diagnostic screen on configuration error

deepakshirkem:
Closes: NA
Related: https://github.com/canonical/mir/issues/3637

What's New?

When Ubuntu Frame encounters a configuration error on startup, Instead of dying the error message on screen using Frame's existing diagnostic feature.

Before: Frame exist silently on configuration error, user sees a black screen with no useful information.

After: Frame re-execs with minimal configuration and displays the error on screen.

Screenshot from 2026-03-24 20-28-16


AlanGriffiths (changes requested):
1. This will not provide "safe" configuration from the configuration file (nor environment variables). The configuration file is where most of Frame's configuration is provided.
2. The diagnostic display will not be seen if we allow clients to connect as they will be displayed in preference.

To avoid parsing the (probably probematic) configuration file it is necessary to construct MirRunner with an explict, safe, config option.

To avoid clients connecting, it is probably easiest to set `WAYLAND_DISPLAY` in the environment: `WAYLAND_DISPLAY=wayland-66` or similar.

src/safe_mode.cpp
AlanGriffiths: All source files require a copyright/licensing header comment.

AlanGriffiths (changes requested):

AlanGriffiths (changes requested):
The configuration file is where most of Frame's configuration is provided. This will still be parsed on reexec. To avoid parsing the (probably problematic) configuration file it is necessary to construct MirRunner with an explicit, safe, config option.

AlanGriffiths:
> Hi @AlanGriffiths, If I understand correctly, the issue is that when Frame re-execs in safe mode, `MIrRunner` still reads the sane broken config file because it uses the default config path. To fix this should Pass `/dev/null `as the config file on re-exec so no config is read at all or Create a minimal safe config file with known good default setting.
> which approach make more sense ?

@deepakshirkem sorry, I missed responding to the above. As I said above:

> To avoid parsing the (probably probematic) configuration file it is necessary to construct MirRunner with an explict, safe, config option.

The `configure` hook provides some default configuration regardless of the snap `config` option, I think that should be the basis of the "known default" setting. (Just tweak the hook to also output these settings to a second file.)

deepakshirkem:
Hi @AlanGriffiths, No worries, thank you for your response! I have addressed your suggestions  please review when you get a chance

src/safe_mode.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+#include <cerrno>
+#include <cstring>
+#include <fstream>
+#include <unistd.h>
+#include <filesystem>
+
+namespace 
+{
+    auto const safe_mode_env = "FRAME_SAFE_MODE";
```
AlanGriffiths: This constant should probably be declared in `safe_mode.h`, not defined in two separate placed.

snap/hooks/configure
```diff
@@ -51,10 +54,25 @@ ${config_option:-# (none)}
...
 config_entry() {
   if ! grep -q "^$1=" "${config_temp}"; then
     cat <<EOT >> "${config_temp}"
 
+# $3
+$1=$2
+EOT
+  fi
+  # also write to safe config
+  if ! grep -q "^$1=" "${safe_config_temp}"; then
```
AlanGriffiths: I don't think we need this test: there should be no user configuration in this file that we need avoid overriding

AlanGriffiths (changes requested):
Sorry (again) for the slow response.

This looks a lot closer, thanks

snap/hooks/configure
deepakshirkem: Removed.

src/safe_mode.cpp
deepakshirkem: Moved

deepakshirkem:
Hi @AlanGriffiths, Please go one more time

AlanGriffiths:
How have you tested this?

It didn't work for me on my first attempt.

1. On my laptop: build & install
2. `sudo snap set ubuntu-frame config="platform-display-libs=nonsense"`
3. `sudo snap start --enable ubuntu-frame`

Expected: diagnostic message
Actual: blank screen (and VT locked up)

I also tried running in a user session:

1. `sudo snap set ubuntu-frame config="display-report=nonsense"`
2. `ubuntu-frame`

Expected: diagnostic message
Actual:
> terminate called after throwing an instance of 'boost::wrapexcept<std::runtime_error>'
>   what():  Exception while creating graphics platform
> Invalid display-report option: nonsense (valid options are: "off" and "log" and "lttng")
> 
> Aborted (core dumped)

deepakshirkem:
Hi @AlanGriffiths,  Sorry for the incomplete testing. I only tested with unknown options and missed the invalid value cases.

### Test 1 — Unknown CLI option:  

```
./frame --invalid-option=garbage
```
### Test 2 — Invalid environment variable:
```
MIR_SERVER_INVALID_KEY=garbage ./frame
```

### Test 3 — Invalid config file key:

```
echo "invalid_key = garbage" > ~/.config/frame.config
./frame
```

I will push an updated commit shortly.

Saviq:
> I will push an updated commit shortly.

Ping! :)

deepakshirkem:
> > I will push an updated commit shortly.
> 
> Ping! :)

Hi @Saviq, Sorry for the delay  I missed this one. I have updated the patch and also verified it with tests. Please check one more time when you get a chance.

AlanGriffiths (changes requested):
Are you testing the snap?

You are stripping `MIR_SERVER_PLATFORM_PATH` (Which, in the Snap environment, needs to be something `"${SNAP}/usr/lib/${CRAFT_ARCH_TRIPLET_BUILD_FOR}/mir/server-platform"` for Mir to start.)

```
$ sudo snap set ubuntu-frame config="display-report=nonsense"
$ ubuntu-frame
...
[2026-06-01 16:52:43.275636] <information> mircommon: Loading modules from: /usr/lib/x86_64-linux-gnu/mir/server-platform
[2026-06-01 16:52:43.275645] < - ERROR - > mircommon: Failed to load libraries from path: /usr/lib/x86_64-linux-gnu/mir/server-platform (error was:/usr/lib/x86_64-linux-gnu/mir/server-platform: No such file or directory)
ERROR: ./src/server/graphics/default_configuration.cpp(200): Throw in function virtual const std::vector<std::shared_ptr<mir::graphics::DisplayPlatform> >& mir::DefaultServerConfiguration::the_display_platforms()
std::exception::what: Exception while creating graphics platform
ERROR: Dynamic exception type: std::system_error
std::exception::what: /usr/lib/x86_64-linux-gnu/mir/server-platform: No such file or directory
```

deepakshirkem:
> Are you testing the snap?
> 
> You are stripping `MIR_SERVER_PLATFORM_PATH` (Which, in the Snap environment, needs to be something `"${SNAP}/usr/lib/${CRAFT_ARCH_TRIPLET_BUILD_FOR}/mir/server-platform"` for Mir to start.)
> 
> ```
> $ sudo snap set ubuntu-frame config="display-report=nonsense"
> $ ubuntu-frame
> ...
> [2026-06-01 16:52:43.275636] <information> mircommon: Loading modules from: /usr/lib/x86_64-linux-gnu/mir/server-platform
> [2026-06-01 16:52:43.275645] < - ERROR - > mircommon: Failed to load libraries from path: /usr/lib/x86_64-linux-gnu/mir/server-platform (error was:/usr/lib/x86_64-linux-gnu/mir/server-platform: No such file or directory)
> ERROR: ./src/server/graphics/default_configuration.cpp(200): Throw in function virtual const std::vector<std::shared_ptr<mir::graphics::DisplayPlatform> >& mir::DefaultServerConfiguration::the_display_platforms()
> Dynamic exception type: boost::wrapexcept<std::runtime_error>
> std::exception::what: Exception while creating graphics platform
> ERROR: Dynamic exception type: std::system_error
> std::exception::what: /usr/lib/x86_64-linux-gnu/mir/server-platform: No such file or directory
> ```

Hi @AlanGriffiths, Very sorry. I was previously testing with the local binary instead of the snap, which is why I missed the MIR_SERVER_PLATFORM_PATH issue.  I have now tested with the snap and fixed.

deepakshirkem:
Ping @AlanGriffiths

src/safe_mode.cpp
```diff
@@ -0,0 +1,128 @@
...
+{
+    void reexec_in_safe_mode(std::string const& error_message, std::string const& diagnostic_path, char const* argv[])
+    {
+        bool diagnostic_written = false;
+        if(std::ofstream diagnostic{diagnostic_path})
+        {
+            diagnostic << "Ubuntu Frame is running in safe mode after error:\n\n"
+                    << error_message << "\n\n"
+                    << "Please contact support";
+            diagnostic_written = true;
```
AlanGriffiths: Is this the best information we can give the end user? If we think of, say, a web kiosk then someone trying to use it will never have heard of "Ubuntu Frame".

AlanGriffiths (approved):
The basic functionality now works.

However, I wonder if we can improve the messaging. The existing diagnostic facility allows the diagnostic screen content to be configured. That allows, for example, supplying details of how to "contact support".

Obviously, this is a fallback for when problems, such as configuration errors, which means this customization opportunities may be limited. But identifying "Ubuntu Frame" may lead to confusion. Imagine yourself as an end-user encountering this, struggling to find a forum for Ubuntu Frame support and, ultimately, being disappointed by the support available.

(Approving as this is an improvement, just wondering if we can do better.)

src/safe_mode.cpp
deepakshirkem: Hi @AlanGriffiths, Do you think any of these messages would work better for end users?

Message 1:
```
This device encountered a configuration error and is running in safe mode:

<error details>

Please contact the device administrator to resolve this issue.
```

Message 2:
```
Safe mode: This device could not start normally due to a configuration error:

<error details>

To restore normal operation, please correct the configuration and restart.
```

Message 3:

```
Configuration error:

<error details>

Please restore the previous configuration or contact the person who manages this device.
```

AlanGriffiths: Perhaps a mix of 2 & 3?

```
Safe mode: This device could not start normally due to a configuration error:
 
<error details>

To restore normal operation please:
  * correct the configuration; or,
  * contact the person who manages this device.
```

AlanGriffiths (approved):

src/safe_mode.cpp
deepakshirkem: Looks good to me. Thank You ::))

deepakshirkem:
Hi @AlanGriffiths, Does this PR require anything else from my side, or is it ready to be merged?

AlanGriffiths:
> Hi @AlanGriffiths, Does this PR require anything else from my side, or is it ready to be merged?

It is ready, just waiting on infrastructure doing the right thing

PR #306, opened 2026-04-27
Add mini accessibility launcher

tarek-y-ismail:
## What's new

- Add's a mini accessibility launcher/widget that communicates with Ubuntu Frame via a config override file

## How to test

First, you need to get a version of Mir with support for live config override, you can use the code corresponding to this PR: https://github.com/canonical/mir/pull/4781

To build a local copy, run the following bash commands:
```
cmake -B build -DCMAKE_INSTALL_PREFIX=~/mir-local-install
cmake --build build -j$(nproc)
cmake --install build
```

Second, build Ubuntu frame from this PR with the following commands. This will use the version you just built and installed:
```
PKG_CONFIG_PATH=~/mir-local-install/lib/pkgconfig:~/mir-local-install/lib/x86_64-linux-gnu/pkgconfig  \
  cmake -B src/build src/
cmake --build src/build -j$(nproc)
```
(Tell me if there's a better way)

Then, run Ubuntu Frame and the launcher:
```
# Terminal 1 – start your locally built ubuntu-frame binary
WAYLAND_DISPLAY=wayland-98 /path/to/ubuntu-frame/src/build/frame \
  --add-wayland-extensions zwlr_layer_shell_v1:zwlr_foreign_toplevel_manager_v1

# Terminal 2 – run the launcher against the same socket
cd /path/to/ubuntu-frame/launcher
UBUNTU_FRAME_LAUNCHER_ACCESSIBILITY_CONFIG_PATH=~/.config/mir/accessibility.ini.d/launcher-override.conf WAYLAND_DISPLAY=wayland-98 flutter run
```

Now, you should have the Ubuntu Frame with the launcher inside of it. Tapping, or pressing and holding options should cycle forwards and backwards, with effects applying immediately.

You can additionally pass an environment variable `UBUNTU_FRAME_LAUNCHER_ACCESSIBILITY_OPTIONS=<option>,<another option>,...` to constrain the options available to users. By default, if the variable is empty, then no options are shown. If it's not present, all options are shown.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -394,6 +394,7 @@ intersphinx_mapping = {
     "frame-22": ("https://ubuntu.com/frame/docs/22", None),
     "server": ("https://ubuntu.com/server/docs", None),
     "snapcraft": ("https://documentation.ubuntu.com/snapcraft/stable", None),
+    "snap": ("https://snapcraft.io/docs/", None),
     "subiquity": (
         "https://canonical-subiquity.readthedocs-hosted.com/en/latest",
         None,
--- a/launcher/lib/controllers/accessibility_controller.dart
+++ b/launcher/lib/controllers/accessibility_controller.dart
@@ -24,7 +24,7 @@ import 'package:ubuntu_frame_launcher/models/accessibility_option.dart';
 class AccessibilityController {
   static const optionsEnvKey = 'UBUNTU_FRAME_LAUNCHER_ACCESSIBILITY_OPTIONS';
   static const _accessibilityConfigKey =
-      "UBUNTU_FRAME_LAUNCHER_ACCESSIBILITY_CONFIG_PATH";
+      "UBUNTU_FRAME_ACCESSIBILITY_CONFIG_PATH";
   static final allOptions = [
     AccessibilityOption(
       id: 'magnifier',
@@ -122,7 +122,7 @@ class AccessibilityController {
   }
 
   /// The path at which the accessibility INI file is written. Reads the
-  /// [UBUNTU_FRAME_LAUNCHER_ACCESSIBILITY_CONFIG_PATH] environment variable;
+  /// [UBUNTU_FRAME_ACCESSIBILITY_CONFIG_PATH] environment variable;
   /// returns an empty string (disabling file I/O) if the variable is not set.
   static String get _configPath =>
       Platform.environment[_accessibilityConfigKey] ?? '';
--- a/launcher/lib/controllers/accessibility_controller.dart
+++ b/launcher/lib/controllers/accessibility_controller.dart
@@ -81,47 +81,32 @@ class AccessibilityController {
             .info('Option "${option.id}" not found in config; keeping default');
         continue;
       }
-      final applied = option.setValueFromString(stored);
-      if (applied) {
-        _logger
-            .info('Option "${option.id}" initialised to "$stored" from config');
-      } else {
-        _logger.warning(
-            'Option "${option.id}" stored value "$stored" is not in the '
-            'allowed values list ${option.values}; keeping default');
-      }
+      option.setValueFromString(stored);
     }
   }
 
   void expand() {
-    _logger.info('Accessibility panel expanded');
     _isExpanded = true;
     _controller.add(true);
   }
 
   void collapse() {
-    _logger.info('Accessibility panel collapsed');
     _isExpanded = false;
     _controller.add(false);
   }
 
   void cycleForward(AccessibilityOption option) {
     option.cycleForward();
-    _logger.info(
-        'Option "${option.label}" cycled forward -> "${option.currentValue}"');
     _writeCurrentConfig();
   }
 
   void cycleBackward(AccessibilityOption option) {
     option.cycleBackward();
-    _logger.info(
-        'Option "${option.label}" cycled backward -> "${option.currentValue}"');
     _writeCurrentConfig();
   }
 
   Future<void> _writeCurrentConfig() async {
     if (_configPath.isEmpty) {
-      _logger.warning('ACCESSIBILITY_CONFIG_PATH is not set; skipping write');
       return;
     }
 
@@ -134,7 +119,6 @@ class AccessibilityController {
       final file = File(_configPath);
       await file.parent.create(recursive: true);
       await file.writeAsString(buffer.toString());
-      _logger.info('Wrote accessibility config to $_configPath');
     } catch (e) {
       _logger.shout('Failed to write accessibility config to $_configPath: $e');
     }
--- a/launcher/lib/controllers/accessibility_controller.dart
+++ b/launcher/lib/controllers/accessibility_controller.dart
@@ -1,0 +1,186 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import 'dart:io';
+
+import 'package:flutter/material.dart';
+import 'package:ini/ini.dart';
+import 'package:logging/logging.dart';
+import 'package:ubuntu_frame_launcher/models/accessibility_option.dart';
+
+class AccessibilityController {
+  static const optionsEnvKey = 'UBUNTU_FRAME_LAUNCHER_ACCESSIBILITY_OPTIONS';
+  static const _accessibilityConfigKey =
+      "UBUNTU_FRAME_LAUNCHER_ACCESSIBILITY_CONFIG_PATH";
+  static final allOptions = [
+    AccessibilityOption(
+      id: 'magnifier',
+      configKey: 'magnifier_enable',
+      icon: Icons.zoom_in,
+      values: ['true', 'false'],
+    ),
+    AccessibilityOption(
+      id: 'output-filter',
+      configKey: 'output_filter',
+      icon: Icons.filter_b_and_w,
+      values: ['none', 'grayscale', 'invert'],
+    ),
+    AccessibilityOption(
+      id: 'cursor-scale',
+      configKey: 'cursor_scale',
+      icon: Icons.ads_click_outlined,
+      values: ['1', '1.5', '2'],
+    ),
+  ];
+
+  static final _logger = Logger('AccessibilityController');
+  final List<AccessibilityOption> options;
+
+  AccessibilityController._({required this.options});
+
+  static List<AccessibilityOption> _getActiveOptions() {
+    final optionsEnv = Platform.environment[optionsEnvKey];
+
+    // Not set, show all
+    if (optionsEnv == null) {
+      return allOptions;
+    }
+
+    // Set to be explicitly empty
+    if (optionsEnv.isEmpty) {
+      return [];
+    }
+
+    final activeOptions = <AccessibilityOption>[];
+    for (final id in optionsEnv.split(',')) {
+      final optionIndex = allOptions.indexWhere((o) => o.id == id);
+      if (optionIndex != -1) {
+        activeOptions.add(allOptions[optionIndex]);
+      } else {
+        _logger.warning('$optionsEnvKey: unknown option "$id", skipping');
+      }
+    }
+
+    return activeOptions;
+  }
+
+  static Future<List<AccessibilityOption>> _loadActiveOptionValues(
+      List<AccessibilityOption> activeOptions, String configPath) async {
+    if (configPath.isEmpty) {
+      _logger.warning(
+          '$_accessibilityConfigKey is not set; skipping startup read');
+      return activeOptions;
+    }
+
+    final file = File(configPath);
+    if (!await file.exists()) {
+      _logger.info('No accessibility config file found at $configPath; '
+          'all options will use their defaults');
+      return activeOptions;
+    }
+
+    final Map<String, String> storedValues = {};
+    try {
+      final contents = await file.readAsString();
+      final config = Config.fromString(contents);
+      // The ini parsing library we use puts everything not under a section into
+      // the 'default' section. It will always exist.
+      const section = 'default';
+      for (final key in config.options(section)!) {
+        storedValues[key] = config.get(section, key)!;
+      }
+    } catch (e, stackTrace) {
+      _logger.shout('Failed to read accessibility config from $configPath', e,
+          stackTrace);
+      return activeOptions;
+    }
+
+    for (final option in activeOptions) {
+      final stored = storedValues[option.configKey];
+      if (stored == null) {
+        _logger
+            .info('Option "${option.id}" not found in config; keeping default');
+        continue;
+      }
+      option.setValueFromString(stored);
+    }
+
+    return activeOptions;
+  }
+
+  /// The path at which the accessibility INI file is written. Reads the
+  /// [UBUNTU_FRAME_LAUNCHER_ACCESSIBILITY_CONFIG_PATH] environment variable;
+  /// returns an empty string (disabling file I/O) if the variable is not set.
+  static String get _configPath =>
+      Platform.environment[_accessibilityConfigKey] ?? '';
+
+  /// Reads the accessibility config file on disk and sets each option's
+  /// current value. Options absent from the file, or with an unrecognised
+  /// value, are left at their default (index 0).
+  static Future<AccessibilityController> create() async {
+    final activeOptions = _loadActiveOptionValues(
+        AccessibilityController._getActiveOptions(), _configPath);
+    return AccessibilityController._(options: await activeOptions);
+  }
+
+  void cycleForward(AccessibilityOption option) {
+    option.cycleForward();
+    _enqueueWrite();
+  }
+
+  void cycleBackward(AccessibilityOption option) {
+    option.cycleBackward();
+    _enqueueWrite();
+  }
+
+  /// Pending write chain — each write is appended here so they are
+  /// executed strictly in order and never overlap.
+  Future<void> _writeFuture = Future.value();
+
+  /// Snapshot the current option values and append an atomic write to the
+  /// serial queue. Any number of rapid calls will be ordered correctly
+  /// because each write is chained onto the previous one.
+  void _enqueueWrite() {
+    // Capture the values *now*, before any await, so the write reflects
+    // the state at the moment this cycle happened.
+    final snapshot = {for (final opt in options) opt.configKey: opt.currentValue};
+    _writeFuture = _writeFuture.then((_) => _writeSnapshot(snapshot));
+  }
+
+  Future<void> _writeSnapshot(Map<String, String> snapshot) async {
+    if (_configPath.isEmpty) return;
+
+    final buffer = StringBuffer();
+    for (final entry in snapshot.entries) {
+      buffer.writeln('${entry.key} = ${entry.value}');
+    }
+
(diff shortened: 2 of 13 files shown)


launcher/lib/controllers/accessibility_controller.dart
tarek-y-ismail, line 1: Maybe too much logging

launcher/lib/controllers/accessibility_controller.dart
```diff
@@ -0,0 +1,142 @@
...
+  void cycleBackward(AccessibilityOption option) {
+    option.cycleBackward();
+    _logger.info(
+        'Option "${option.label}" cycled backward -> "${option.currentValue}"');
+    _writeCurrentConfig();
+  }
+
+  Future<void> _writeCurrentConfig() async {
+    if (_configPath.isEmpty) {
+      _logger.warning('ACCESSIBILITY_CONFIG_PATH is not set; skipping write');
```
tarek-y-ismail: Should use `_accessibilityConfigKey`

src/frame_main.cpp
```diff
@@ -44,6 +53,32 @@ int main(int argc, char const* argv[])
...
+    miral::CursorScale cursor_scale{config_aggregator};
+    miral::OutputFilter output_filter{config_aggregator};
+    miral::Magnifier magnifier{config_aggregator};
+
+    miral::ConfigFile config_file{
+        runner,
+        "/home/tarek.ismail@canonical.com/.config/mir/accessibility.ini",
+        miral::ConfigFile::Mode::reload_on_change,
+        [&adapter](auto args) { adapter(args); },
+    };
```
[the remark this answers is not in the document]
Saviq: This!

Saviq (changes requested):
Remember we want to expose a `launcher.{head,body,tail}=[...]` configuration.

For now it'd be enough that we support `.tail=[magnifier,output_filter,cursor_scale]`, so we need to be able to pass that list in (environment variable is fine) and add the icons in the order given.

Copilot has some thoughts, too.

launcher/lib/controllers/accessibility_controller.dart
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    final Map<String, String> storedValues = {};
+    try {
+      final contents = await file.readAsString();
+      final config = Config.fromString(contents);
+      const section = 'default';
+      for (final key in config.options(section)!) {
+        storedValues[key] = config.get(section, key)!;
+      }
```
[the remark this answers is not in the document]
tarek-y-ismail: The library we use to read INI files puts any keys/values not already under a section into `[default]`. So this is a non-issue.

src/frame_main.cpp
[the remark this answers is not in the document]
tarek-y-ismail: Non-issue for now. We can handle it when it pops up.

src/frame_main.cpp
Saviq, line 1: We'll want macro guards in here to be able to build with older Mir.

src/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ project(ubuntu-frame)
 include(FindPkgConfig)
 include(CMakeDependentOption)
 
-set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD 23)
```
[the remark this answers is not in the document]
Saviq: Need to bump `cmake_minimum_required()` for a better diagnostic:

https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html

Saviq (changes requested):
Should we drop the expanding "drawer", and just keep the selected options on screen all the time?

And I think we can drop the labels and tooltips, they're not very accessible ;)

Keep it simple!

launcher/lib/main.dart
```diff
@@ -39,13 +47,54 @@ void main() async {
...
+      icon: Icons.filter_b_and_w,
+      values: ['none', 'grayscale', 'invert'],
+    ),
+    'cursor_scale': AccessibilityOption(
+      id: 'cursor_scale',
+      label: 'Cursor Scale',
+      icon: Icons.mouse_outlined,
+      values: ['1', '1.5', '2'],
+    ),
+  };
```
mattkae: The map here feels wrong. We already have an `id` that matches the key in the map. Can we not just have a `List`?

launcher/lib/main.dart
```diff
@@ -39,13 +47,54 @@ void main() async {
...
+      } else {
+        logger.warning('$optionsEnvKey: unknown option "$id", skipping');
+      }
+    }
+  } else {
+    activeOptions = allOptions.values.toList();
+  }
+
+  getIt.registerLazySingleton<AccessibilityController>(
+      () => AccessibilityController(options: activeOptions));
```
mattkae: I would move the logic to calculate `activeOptions` into the controller constructor itself. It will localize the implementation, which is quite involved.

launcher/lib/views/dock.dart
```diff
@@ -17,39 +17,53 @@
...
+import 'package:ubuntu_frame_launcher/views/accessibility_button.dart';
 import 'package:ubuntu_frame_launcher/views/dock_button.dart';
 import 'package:ubuntu_frame_launcher/views/stream_builder_with_future_initial_value.dart';
 
 const double _dockWidthPx = 70;
 const _dockPadding = EdgeInsets.fromLTRB(3, 6, 3, 6);
 
-class Dock extends StatelessWidget {
-  final applicationControler = GetIt.instance.get<ApplicationController>();
+class Dock extends StatefulWidget {
```
mattkae: This should remain `Stateless`, since we don't use any state :)

launcher/lib/models/accessibility_option.dart
```diff
@@ -0,0 +1,64 @@
...
+    required this.values,
+    int currentIndex = 0,
+  })  : assert(values.isNotEmpty),
+        assert(currentIndex >= 0 && currentIndex < values.length),
+        _currentIndex = currentIndex.clamp(0, values.length - 1);
+
+  int get currentIndex => _currentIndex;
+  String get currentValue => values[_currentIndex];
+
+  Stream<String> getStream() => _controller.stream;
```
mattkae:
```suggestion
  Stream<String> get stream => _controller.stream;
```

launcher/lib/controllers/accessibility_controller.dart
```diff
@@ -0,0 +1,145 @@
...
+
+  final _controller = StreamController<bool>.broadcast();
+
+  AccessibilityController({required this.options});
+
+  /// Pending write chain — each write is appended here so they are
+  /// executed strictly in order and never overlap.
+  Future<void> _writeFuture = Future.value();
+
+  Stream<bool> getStream() => _controller.stream;
```
mattkae:
```suggestion
  Stream<bool> get stream => _controller.stream;
```

launcher/lib/controllers/accessibility_controller.dart
```diff
@@ -0,0 +1,145 @@
...
+
+  /// The path at which the accessibility INI file is written. Reads the
+  /// [UBUNTU_FRAME_LAUNCHER_ACCESSIBILITY_CONFIG_PATH] environment variable;
+  /// returns an empty string (disabling file I/O) if the variable is not set.
+  String get _configPath => Platform.environment[_accessibilityConfigKey] ?? '';
+
+  /// Reads the accessibility config file on disk and sets each option's
+  /// current value. Options absent from the file, or with an unrecognised
+  /// value, are left at their default (index 0).
+  Future<void> initialize() async {
```
mattkae: Instead of having a two-step initialization, a better solution is to use a static creator method.

Essentially, make the main constructor be private `using `AccessibilityController._(...)` and instead expose an async builder:

```dart
static Future<AccessibilityController> create(...) { }
```

That way, creation can be awaited _and_ its guaranteed to happen without the need for two calls.

mattkae (dismissed):
Some minor points, but looks pretty snazzy overall!

tarek-y-ismail:
@Saviq I feel like there's too little feedback right now. Try it out and tell me what you think.

Saviq:
> I feel like there's too little feedback right now.

I suppose we just need an "enabled" state for the buttons, but also most of the feedback will actually happen on screen, no?

FWIW I can't this to act in Frame built from canonical/mir#4781… it doesn't seem to be reacting to the drop-in file changing. I see nothing in the logs about the a11y config file, either.

tarek-y-ismail:
This has two sides:
1. Configuring the Mir side: Mir expects the configuration file to be in `{$XDG_HOME,$HOME}/.config/mir/accessibility.ini` (come to think of it, this may need some more configuration?)
2. Configuring on the launcher side: `UBUNTU_FRAME_LAUNCHER_ACCESSIBILITY_CONFIG_PATH={$XDG_HOME,$HOME}/.config/mir/accessibility.ini.d/*.conf` 

You can troubleshoot the first side by checking the logs, I print out all the files being loaded on every reload. First check if the file is loaded during the initial launch, then modify and save it to check if it's being reloaded. For the second step, make sure the launcher is writing to the proper file, and then make sure it's being picked up by Mir.

tarek-y-ismail:
> but also most of the feedback will actually happen on screen, no?

For the ones we have right now, yes. If we expand to keyboard options, then those won't be immediately visible. I suppose we can reintroduce labels then if it ever happens.

Saviq:
> FWIW I can't this to act in Frame

NVM, had bad environment, works now!

launcher/lib/controllers/accessibility_controller.dart
```diff
@@ -0,0 +1,181 @@
...
+  static final _logger = Logger('AccessibilityController');
+  final List<AccessibilityOption> options;
+
+  AccessibilityController._({required this.options});
+
+  static List<AccessibilityOption> _getActiveOptions() {
+    final optionsEnv = Platform.environment[optionsEnvKey];
+
+    final List<AccessibilityOption> activeOptions;
+    if (optionsEnv != null && optionsEnv.isNotEmpty) {
```
Saviq: We need to be able to disable it, at least if the environment option is set, but empty, we should have no a11y items on the launcher.

If it's unset, the default might indeed be all of them.

launcher/lib/controllers/accessibility_controller.dart
```diff
@@ -0,0 +1,181 @@
...
+      activeOptions = [];
+      for (final id in optionsEnv.split(':')) {
```
Saviq:
```suggestion
      for (final id in optionsEnv.split(',')) {
```

launcher/lib/controllers/accessibility_controller.dart
```diff
@@ -0,0 +1,181 @@
...
+        if (optionIndex != -1) {
+          activeOptions.add(allOptions[optionIndex]);
+        } else {
+          _logger.warning('$optionsEnvKey: unknown option "$id", skipping');
+        }
+      }
+      if (activeOptions.isEmpty) {
+        _logger.warning(
+            '$optionsEnvKey: no valid options found in "$optionsEnv", showing all options');
+        return allOptions;
```
Saviq:
```suggestion
```

launcher/lib/controllers/accessibility_controller.dart
```diff
@@ -0,0 +1,181 @@
...
+import 'package:logging/logging.dart';
+import 'package:ubuntu_frame_launcher/models/accessibility_option.dart';
+
+class AccessibilityController {
+  static final optionsEnvKey = 'UBUNTU_FRAME_LAUNCHER_ACCESSIBILITY_OPTIONS';
+  static final _accessibilityConfigKey =
+      "UBUNTU_FRAME_LAUNCHER_ACCESSIBILITY_CONFIG_PATH";
+  static final allOptions = [
+    AccessibilityOption(
+      id: 'magnifier_enable',
```
Saviq: Let's not limit the functionality of this to just enabling.

```suggestion
      id: 'magnifier',
```

tarek-y-ismail: `id` has to match the option name Mir is expecting.

launcher/lib/controllers/accessibility_controller.dart
```diff
@@ -0,0 +1,181 @@
...
+      values: ['true', 'false'],
+    ),
+    AccessibilityOption(
+      id: 'output_filter',
+      icon: Icons.filter_b_and_w,
+      values: ['none', 'grayscale', 'invert'],
+    ),
+    AccessibilityOption(
+      id: 'cursor_scale',
+      icon: Icons.mouse_outlined,
```
Saviq: Maybe this?

```suggestion
      icon: Icons.ads_click_outlined,
```

launcher/lib/controllers/accessibility_controller.dart
Saviq: It shouldn't _have_ to, that's the thing.

We're exposing a new `snap set …` interface to the user. What's behind the scenes is an implementation detail.

EDIT: we might also want to introduce changing the size and/or zoom level of the magniifier, don't want to have another option for that.

launcher/lib/controllers/accessibility_controller.dart
```diff
@@ -0,0 +1,182 @@
...
+  static final _accessibilityConfigKey =
+      "UBUNTU_FRAME_LAUNCHER_ACCESSIBILITY_CONFIG_PATH";
+  static final allOptions = [
+    AccessibilityOption(
+      id: 'magnifier_enable',
+      icon: Icons.zoom_in,
+      values: ['true', 'false'],
+    ),
+    AccessibilityOption(
+      id: 'output_filter',
```
Saviq:
```suggestion
      id: 'output-filter',
```

launcher/lib/controllers/accessibility_controller.dart
```diff
@@ -0,0 +1,182 @@
...
+      icon: Icons.filter_b_and_w,
+      values: ['none', 'grayscale', 'invert'],
+    ),
+    AccessibilityOption(
+      id: 'cursor_scale',
```
Saviq:
```suggestion
      id: 'cursor-scale',
```

Saviq (dismissed):
Hyphens are nicer to humans.

launcher/lib/controllers/accessibility_controller.dart
Saviq: We could also translate it in the wrapper, but this is probably a better place to do so.

@copilot please add another field for `config_key` so that option IDs can map to another key in the override file.

launcher/lib/controllers/accessibility_controller.dart
tarek-y-ismail: These names are not visible to users. And as mentioned before, they must match the option name on mir's end.

launcher/lib/controllers/accessibility_controller.dart
Saviq: @copilot drop the default treatment of `configKey`, make it required instead.

launcher/lib/controllers/accessibility_controller.dart
```diff
@@ -0,0 +1,185 @@
...
+
+import 'dart:io';
+
+import 'package:flutter/material.dart';
+import 'package:ini/ini.dart';
+import 'package:logging/logging.dart';
+import 'package:ubuntu_frame_launcher/models/accessibility_option.dart';
+
+class AccessibilityController {
+  static final optionsEnvKey = 'UBUNTU_FRAME_LAUNCHER_ACCESSIBILITY_OPTIONS';
```
robert-ancell: `static final` -> `static const`

launcher/lib/controllers/accessibility_controller.dart
```diff
@@ -0,0 +1,185 @@
...
+    if (optionsEnv == null) {
+      return allOptions;
+    }
+
+    // Set to be explicitly empty
+    if (optionsEnv.isEmpty) {
+      return [];
+    }
+
+    List<AccessibilityOption> activeOptions = [];
```
robert-ancell: More clear as `var activeOptions = <AccessibilityOption>[];`

launcher/lib/controllers/accessibility_controller.dart
```diff
@@ -0,0 +1,185 @@
...
+          'all options will use their defaults');
+      return activeOptions;
+    }
+
+    final Map<String, String> storedValues = {};
+    try {
+      final contents = await file.readAsString();
+      final config = Config.fromString(contents);
+      const section = 'default';
+      for (final key in config.options(section)!) {
```
robert-ancell, line 102: This will crash if the section does not exist, probably safer as:
```dart
config.options(section) ?? []
```

robert-ancell (approved):
Some small style suggestions but otherwise no problems I can see. It's worth running `dart format` and `dart analyze` over this - these pick up some small things. But the existing code doesn't cleanly analyze. Ideally this would be put into the CI.

robert-ancell:
Note: I reviewed the Dart/Flutter code - please get the other reviewers approval before landing!

launcher/lib/controllers/accessibility_controller.dart
tarek-y-ismail: The ini parsing library adds everything not under a section under this one. So it will always exist. Added a comment clarifying that.

tarek-y-ismail:
@Saviq Merge time?

Saviq:
> @Saviq Merge time?


Not until the snap builds (IoW, not until override lands, since it already bumped MirAL above the guarded version).

docs/conf.py
```diff
@@ -386,12 +385,13 @@
...
-    "mir": ("https://canonical-mir.readthedocs-hosted.com/stable", None),
-    "mir-latest": ("https://canonical-mir.readthedocs-hosted.com/latest", None),
+    "mir": ("https://canonical.com/mir/stable", None),
+    "mir-latest": ("https://canonical.com/mir/latest", None),
     "core": ("https://documentation.ubuntu.com/core", None),
-    "frame-22": ("https://canonical-ubuntu-frame-documentation.readthedocs-hosted.com/22", None),
+    "frame-22": ("https://ubuntu.com/frame/docs/22", None),
     "server": ("https://documentation.ubuntu.com/server", None),
     "snapcraft": ("https://documentation.ubuntu.com/snapcraft/stable", None),
+    "snap": ("https://snapcraft.io/docs/", None),
```
mattkae, line 397: Should these changes be in a separate commit? Spurious?

launcher/lib/controllers/accessibility_controller.dart
```diff
@@ -0,0 +1,186 @@
...
+    if (optionsEnv == null) {
+      return allOptions;
+    }
+
+    // Set to be explicitly empty
+    if (optionsEnv.isEmpty) {
+      return [];
+    }
+
+    var activeOptions = <AccessibilityOption>[];
```
mattkae:
```suggestion
    final activeOptions = <AccessibilityOption>[];
```

mattkae: `var` is typically avoided at all costs, unless you're actually reassigning this.

mattkae (approved):
Some minor things, non-blocking

Saviq:
> This branch has conflicts that must be resolved

Saviq:
> This branch has conflicts that must be resolved

tarek-y-ismail:
Will do a quick pass over the commit history tomorrow since it's a bit messy and then merge. You've got until then if you want to request changes.

src/frame_main.cpp
```diff
@@ -50,6 +64,32 @@ int main(int argc, char const* argv[])
     runner.add_stop_callback([&] { background_client.stop(); });
     auto display_config = build_display_configuration(runner);
 
+#if MIRAL_VERSION >= MIR_VERSION_NUMBER(5, 8, 0)
```
tarek-y-ismail: Need to check for miral 6.0.0

src/frame_main.cpp
```diff
@@ -76,6 +116,11 @@ int main(int argc, char const* argv[])
                 window_manager_observer,
                 display_config),
             Keymap{},
-            miral::Decorations::always_csd()
+            miral::Decorations::always_csd(),
+#if MIRAL_VERSION >= MIR_VERSION_NUMBER(5, 8, 0)
```
tarek-y-ismail: Need to check for miral 6.0.0

PR #308, opened 2026-04-28
docs: fix canoncial and ogp url versions

SecondSkoll:
On checking the Mir docs mid-migration, I noticed an issue that I missed.

The canonical link and OGP URL provided lacks the version of the documentation.

The following changes ensure the version obtained from the RTD build (which is the `/{version}/` part of the URL - not necessarily the actual product version) is populated into both.


Saviq:
Merged in c13283b

PR #311, opened 2026-04-28
Add support for the plymouth-replace interface

jhenstridge:
This PR is intended to work together with https://github.com/canonical/snapd/pull/16962 to achieve flicker-free boot purely managed from the snap.

If the snap declaration from the snapd PR is accepted, the `plymouth-replace` interface will be `allow-installation: false`, so we would need a snap declaration to allow the ubuntu-frame snap to be installable. I'm unsure whether we'd want to request auto-connection.

### DIFF
--- a/.github/workflows/snap.yml
+++ b/.github/workflows/snap.yml
@@ -40,6 +40,7 @@ jobs:
     - name: Build and publish the snap
       uses: canonical/actions/build-snap@78e8dd5f693653767f2c60c4a6a421a1ebed35af # release @ 2025.06.05
       with:
+        snapcraft-channel: edge
         architecture: ${{ matrix.platform }}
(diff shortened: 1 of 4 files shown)


tarek-y-ismail:
Will need some time to digest this story, but I'm working on it. Once I'm done I'll start reviewing this PR. If anyone wants to have a go, let me know :)

tarek-y-ismail (approved):
Makes sense. I'll leave it to you to press the button once snapd merges the other PR. The CLA fails because it sees two different emails, should be easy enough to rewrite the history.

jhenstridge:
It's not worth merging this until we are reasonably certain the snapd interface will be accepted.

Also we will need to get the snap-declaration updated in the store to say that ubuntu-frame is allowed to have a `plymouth-replace` plug, since the proposal is to make the interface super privileged.

PR #312, opened 2026-04-28
snap: prefix version with quantifier

Saviq:
The number isn't immediately clear otherwise.


AlanGriffiths (approved):
I had a feeling we've bumped against the 32char limit with CI builds. But I don't see where: `462-mir2.27.0~dev279` seems as long as we get.

So, sure. Or would `frame` be a better prefix?

Saviq:
> So, sure. Or would `frame` be a better prefix?

Yeah, that is better, thanks!

Issue #319, opened 2026-06-03
Where to put 'plugs:   graphics-core22' ?


Dadu042:
A example would be welcome.

---
*Reported from: https://ubuntu.com/frame/docs/22/how-to/use-snap-graphics/*

tarek-y-ismail:
As mentioned in the start of the section in the docs, this block is supposed to be added to a `snapcraft.yaml` file.

<img width="872" height="270" alt="Image" src="https://github.com/user-attachments/assets/091f989b-63d5-4968-8726-3c92257c15f6" />

Here's an example of Frame's `snapcraft.yaml` for core22: https://github.com/canonical/ubuntu-frame/blob/core22/snap/snapcraft.yaml

Dadu042:
I know that this place is in 'snapcraft.yaml', but it is uncommon to have to use 'plugs:' like that, I mean to put it before the part of the app snapped.

Saviq:
@Dadu042 would you be able to contribute to https://github.com/canonical/ubuntu-frame/blob/core24/docs/how-to/use-snap-graphics.md what you'd like to see on that page?

Here's the reference for that in Snapcraft docs:

https://documentation.ubuntu.com/snapcraft/stable/reference/snapcraft-yaml/#content-plug-keys

Dadu042:
Ok, done: https://github.com/canonical/ubuntu-frame/pull/320

PR #320, opened 2026-06-18
Add note for section placement in snap graphics

Dadu042:
Added a note about section placement for 'plugs:' support.

As suggested by Saviq, there: https://github.com/canonical/ubuntu-frame/issues/319

### DIFF
--- a/docs/how-to/use-snap-graphics.md
+++ b/docs/how-to/use-snap-graphics.md
@@ -57,7 +57,8 @@ Please check out the code for your specific extension [here](https://github.com/
        target: $SNAP/gpu-2404
        default-provider: mesa-2404
    ```
-Note: this section must be inserted before 'parts:' and 'apps:'.
+
+   Note: this must be inserted as a {ref}`top-level plugs: entry <snapcraft:project.plugs>`.
 
 {#gpu-2404-x11-layouts}
 
--- a/docs/how-to/use-snap-graphics.md
+++ b/docs/how-to/use-snap-graphics.md
@@ -57,6 +57,7 @@ Please check out the code for your specific extension [here](https://github.com/
        target: $SNAP/gpu-2404
        default-provider: mesa-2404
    ```
+Note: this section must be inserted before 'parts:' and 'apps:'.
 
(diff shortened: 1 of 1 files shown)


docs/how-to/use-snap-graphics.md
```diff
@@ -57,6 +57,7 @@ Please check out the code for your specific extension [here](https://github.com/
        target: $SNAP/gpu-2404
        default-provider: mesa-2404
    ```
+Note: this section must be inserted before 'parts:' and 'apps:'.
```
Saviq: That's not true, YAML doesn't care about order?

The only thing is that it needs to be placed as a top level key?

Saviq (changes requested):

docs/how-to/use-snap-graphics.md
Dadu042: For most snapcrafters (newbies) 'plugs:' is a section that we use/put into 'apps:'. But for this one, as I understand (and tested), it must not be placed into 'apps:' nor 'parts:', otherwise the command 'snapcraft pack' does fail.

AlanGriffiths: Yes, there are two "plugs" keys in the snapcraft grammer: "apps.plugs" and "plugs".

But this note is actively misleading and doesn't address the potential for confusion you describe.

AlanGriffiths (changes requested):

docs/how-to/use-snap-graphics.md
Saviq: Yes, it needs to be placed as a top-level key, but not "before".

https://documentation.ubuntu.com/snapcraft/9.0/reference/snapcraft-yaml/#content-plug-keys

```suggestion

   Note: this must be inserted as a {ref}`top-level plugs: entry <snapcraft:project.plugs>`.
```

Dadu042:
Ok, by writing 'before' it seems to me more obvious for snapcraft makers where to put this 'plugs:' part.

Saviq (approved):

PR #322, opened 2026-06-25
Conditionals to work with Mir PR5049

AlanGriffiths:
This deals with one of the ABI breaks coming in Mir-2.29 (and unblocks https://github.com/canonical/mir/pull/5049)

### DIFF
--- a/src/frame_window_manager.cpp
+++ b/src/frame_window_manager.cpp
@@ -51,7 +51,7 @@ bool can_position_be_overridden(WindowSpecification& spec, WindowInfo const& win
     }
 
     // Only override behavior of windows without a parent
-#if MIRAL_VERSION < MIR_VERSION_NUMBER(6, 0, 0)
+#ifdef MIR_OPTIONAL_VALUE_H_
     if (spec.parent().is_set() ? spec.parent().value().lock() : window_info.parent())
 #else
     if (spec.parent().has_value() ? spec.parent().value().lock() : window_info.parent())
@@ -78,7 +78,7 @@ void apply_fullscreen(WindowSpecification& spec)
 {
     spec.state() = mir_window_state_fullscreen;
 
-#if MIRAL_VERSION < MIR_VERSION_NUMBER(6, 0, 0)
+#ifdef MIR_OPTIONAL_VALUE_H_
     spec.size() = mir::optional_value<Size>{};      // Ignore requested size (if any) when we fullscreen
     spec.top_left() = mir::optional_value<Point>{}; // Ignore requested position (if any) when we fullscreen
 #else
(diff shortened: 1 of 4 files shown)


src/frame_window_manager.cpp
```diff
@@ -40,7 +40,7 @@ namespace
 bool can_position_be_overridden(WindowSpecification& spec, WindowInfo const& window_info)
 {
     // Only override behavior of windows of type normal and freestyle
-    switch (spec.type().is_set() ? spec.type().value() : window_info.type())
+    switch (spec.type().value_or(window_info.type()))
     {
         case mir_window_type_normal:
```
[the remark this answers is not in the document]
AlanGriffiths: Wrong, the `mir::optional_value::value_or()` compiles just fine

src/frame_window_manager.cpp
```diff
@@ -73,8 +77,14 @@ bool can_position_be_overridden(WindowSpecification& spec, WindowInfo const& win
...
 {
     spec.state() = mir_window_state_fullscreen;
+
+#if MIRAL_VERSION < MIR_VERSION_NUMBER(6, 0, 0)
     spec.size() = mir::optional_value<Size>{};      // Ignore requested size (if any) when we fullscreen
     spec.top_left() = mir::optional_value<Point>{}; // Ignore requested position (if any) when we fullscreen
+#else
+    spec.size() = std::nullopt;      // Ignore requested size (if any) when we fullscreen
+    spec.top_left() = std::nullopt; // Ignore requested position (if any) when we fullscreen
+#endif
```
mattkae, line 87: Can we simplify to?

```suggestion
    spec.size() = {};      // Ignore requested size (if any) when we fullscreen
    spec.top_left() = {}; // Ignore requested position (if any) when we fullscreen
```

src/layout_metadata.h
```diff
@@ -35,7 +35,11 @@ class LayoutMetadata
     /// Try to assign the window to a positition and size based on its title and snap name.
     /// \returns true if successfully assigned, otherwise false
     bool try_layout(miral::WindowSpecification& specification,
+#if MIRAL_VERSION < MIR_VERSION_NUMBER(6, 0, 0)
         mir::optional_value<std::string> const& title,
+#else
+        std::optional<std::string> const& title,
+#endif
```
mattkae: @copilot Could this check be wrapped up in a macro, e.g. `MIR_OPTIONAL_BACKWARD_COMPAT` that produces the right type depending on the `MIRAL_VERSION`? That would be useful in avoiding macro `if`s throughout this PR by localizing them in a single place.

mattkae (approved):
Approved, but I wonder if we can do something more clever around the const `#if ... #else` blocks everywhere. Perhaps we can centralize that logic into another macro? I asked copilot about the feasibility

AlanGriffiths:
> Approved, but I wonder if we can do something more clever around the const `#if ... #else` blocks everywhere. Perhaps we can centralize that logic into another macro? I asked copilot about the feasibility

Well, we could use:

```c++
namespace mir { using optional_value = std::optional; }
```

With that the code in Frame can be rewritten to compile with both optionals (as shown). It does, of course, come with some technical debt but...

src/frame_window_manager.cpp
AlanGriffiths: I've taken a different approach in Miriway. (I've found the mir-team is happier with pre-processor than compiler magic, hence the approach here.)

mattkae (approved):
Nice, I like that :)

PR #323, opened 2026-07-17
Migrate to miral KioskWindowManager

AlanGriffiths:
Mir is deprecating MinimalWindowManager, and has a better option in KioskWindowManagerPolicy

### DIFF
--- a/src/frame_window_manager.cpp
+++ b/src/frame_window_manager.cpp
@@ -188,7 +188,7 @@ FrameWindowManagerPolicy::FrameWindowManagerPolicy(
     WindowManagerTools const& tools,
     WindowManagerObserver& window_manager_observer,
     miral::DisplayConfiguration const& display_config)
-    : MinimalWindowManager{tools},
+    : KioskWindowManagerPolicy{tools},
       window_manager_observer{window_manager_observer},
       display_config{display_config}
 {
@@ -273,7 +273,7 @@ void FrameWindowManagerPolicy::handle_layout(
 auto FrameWindowManagerPolicy::place_new_window(ApplicationInfo const& app_info, WindowSpecification const& request)
 -> WindowSpecification
 {
-    WindowSpecification specification = MinimalWindowManager::place_new_window(app_info, request);
+    WindowSpecification specification = KioskWindowManagerPolicy::place_new_window(app_info, request);
     WindowInfo window_info{};
     handle_layout(specification, app_info.application(), window_info);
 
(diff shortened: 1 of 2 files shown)


src/frame_main.cpp
tarek-y-ismail, line 1: I would say that the changes in this file should go into another, separate PR.

src/frame_window_manager.cpp
```diff
@@ -230,12 +250,12 @@ void FrameWindowManagerPolicy::handle_layout(
             {
                 mir::log_info(R"(Surface for snap="%s" with title="%s")",
                               snap_instance_name.c_str(),
-                              specification.name().value().c_str());
+                              specification.name().transform([](auto const& name) { return name.c_str(); }).value_or(""));
```
tarek-y-ismail, line 255: We already check if `specification.name()` has a value or not a few lines before this. No need to go through `transform` or `value_or`.

src/frame_window_manager.cpp
```diff
@@ -230,12 +250,12 @@ void FrameWindowManagerPolicy::handle_layout(
...
             }
             else
             {
                 mir::log_info("Surface with title=\"%s\"",
-                              specification.name().value().c_str());
+                              specification.name().transform([](auto const& name) { return name.c_str(); }).value_or(""));
```
tarek-y-ismail, line 260: Same here.

tarek-y-ismail (changes requested):
Some minor nits

src/frame_main.cpp
AlanGriffiths: The "fix TICS build" and "fix code flagged by TICS" commits could indeed be a separate PR. Is it worth it?

tarek-y-ismail (approved):

src/frame_window_manager.cpp
```diff
@@ -118,6 +119,7 @@ auto get_rect_by_force(WindowSpecification const& spec) -> Rectangle
 #else
     mir::fatal_error("WindowSpecification must have both top_left and size set to get a rectangle");
 #endif
+    std::unreachable();
```
tarek-y-ismail, line 122: Quick question before merging, is this intentional or a leftover from the split?

AlanGriffiths: Copilot suggested it to emphasize that fatal error is `[[noreturn]]`

AlanGriffiths: But please land the `fix-tics` PR first

PR #326, opened 2026-07-21
fix TICS build

AlanGriffiths:
FIxes the TICS build, then fixes the code flagged by TICS

### DIFF
--- a/src/frame_window_manager.cpp
+++ b/src/frame_window_manager.cpp
@@ -19,6 +19,7 @@
 #include "snap_name_of.h"
 
 #include <mir/log.h>
+#include <mir/fatal.h>
 #include <miral/application_info.h>
 #include <miral/output.h>
 #include <miral/toolkit_event.h>
@@ -54,7 +55,7 @@ bool can_position_be_overridden(WindowSpecification& spec, WindowInfo const& win
 #ifdef MIR_OPTIONAL_VALUE_H_
     if (spec.parent().is_set() ? spec.parent().value().lock() : window_info.parent())
 #else
-    if (spec.parent().has_value() ? spec.parent().value().lock() : window_info.parent())
+    if (spec.parent().transform([](auto const& parent) { return parent.lock(); }).value_or(window_info.parent()))
 #endif
     {
         return false;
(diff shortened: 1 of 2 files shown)


src/frame_window_manager.cpp
```diff
@@ -230,12 +250,12 @@ void FrameWindowManagerPolicy::handle_layout(
             {
                 mir::log_info(R"(Surface for snap="%s" with title="%s")",
                               snap_instance_name.c_str(),
-                              specification.name().value().c_str());
+                              specification.name().transform([](auto const& name) { return name.c_str(); }).value_or(""));
```
tarek-y-ismail, line 253: `specification.name()` is guaranteed to be valid in this codepath.
```suggestion
                              specification.name()->c_str());
```

src/frame_window_manager.cpp
```diff
@@ -230,12 +250,12 @@ void FrameWindowManagerPolicy::handle_layout(
...
             }
             else
             {
                 mir::log_info("Surface with title=\"%s\"",
-                              specification.name().value().c_str());
+                              specification.name().transform([](auto const& name) { return name.c_str(); }).value_or(""));
```
tarek-y-ismail, line 258: Same here.
```suggestion
                              specification.name()->c_str());
```

tarek-y-ismail (changes requested):
Small carryover nits from the other PR.

src/frame_window_manager.cpp
AlanGriffiths: TICS complains because it isn't guaranteed - as explained in the other PR

tarek-y-ismail: Seems like I missed it. Sorry!

tarek-y-ismail (approved):
Let's land it

tarek-y-ismail:
Looks like "Code scanning results / CodeQL" is blocking merging

AlanGriffiths:
> Looks like "Code scanning results / CodeQL" is blocking merging

No, that's not "required". The "Snap (armhf)" is blocking

Issue #321, opened 2026-06-24
A core26-based snap cannot be installed on Ubuntu Desktop/Server 26.04.


rickwu666666:
I am trying to install ubuntu-frame from the 26/edge channel on Ubuntu Desktop/Server 26.04, but the installation fails with the following error.
```
ubuntu@ubuntu:~$ sudo snap install ubuntu-frame --channel=26/edge
error: cannot perform the following tasks:
- Run configure hook of "ubuntu-frame" snap if present (run hook "configure": 
-----
diff: /var/snap/ubuntu-frame/20799/frame.config: No such file or directory
mv: cannot stat '/var/snap/ubuntu-frame/20799/frame.config': No such file or directory
error: snapctl: snap "ubuntu-frame" option "launcher" is not a map
-----)
```
Additionally, I observed several AppArmor denial logs during the snap installation.
```
ubuntu@ubuntu:~$ sudo journalctl --output=short --follow --all | sudo snappy-debug
kernel.printk_ratelimit = 0
= AppArmor =
Time: Jun 24 14:11:42
Log: apparmor="DENIED" operation="open" class="file" profile="snap.snappy-debug.snappy-debug" name="/proc/28341/mountinfo" pid=28341 comm="snap-exec" requested_mask="r" denied_mask="r" fsuid=0 ouid=0
/proc/28341/mountinfo (read)
Suggestions:
* adjust program to not access '@{PROC}/@{pid}/mountinfo'
* add one of 'mount-control, mount-observe, steam-support' to 'plugs'

= AppArmor =
Time: Jun 24 14:13:02
Log: apparmor="DENIED" operation="open" class="file" profile="snap.ubuntu-frame.hook.install" name="/proc/29463/mountinfo" pid=29463 comm="snap-exec" requested_mask="r" denied_mask="r" fsuid=0 ouid=0
/proc/29463/mountinfo (read)
Suggestions:
* adjust program to not access '@{PROC}/@{pid}/mountinfo'
* add one of 'mount-control, mount-observe, steam-support' to 'plugs'

= AppArmor =
Time: Jun 24 14:13:02
Log: apparmor="DENIED" operation="open" class="file" profile="snap.ubuntu-frame.hook.install" name="/proc/29491/mountinfo" pid=29491 comm="snapctl" requested_mask="r" denied_mask="r" fsuid=0 ouid=0
/proc/29491/mountinfo (read)
Suggestions:
[...]

PR #330, opened 2026-08-18
Only process `launcher.tail` when launcher is a JSON object

AlanGriffiths:
Fixes: #321


Saviq (approved):
ACK as a quick fix, but we really want to get rid of `true`/`false` for 26 and require it to be a JSON object that, if empty, disables the launcher.

PR #333, opened 2026-08-21
launcher: add support for head/tail options

saviq-work:
Deprecate `true`/`false` values for the launcher and require valid JSON objects instead.

---

To test, use the newly introduced workshop:

```
workshop run \
  --env UBUNTU_FRAME_LAUNCHER_TAIL_ITEMS_JSON='["running","magnifier"]' \
  --env UBUNTU_FRAME_LAUNCHER_HEAD_ITEMS_JSON='["running","cursor-scale"]' \
  frame
```

### DIFF
--- a/docs/reference/ubuntu-frame-configuration-options.md
+++ b/docs/reference/ubuntu-frame-configuration-options.md
@@ -58,16 +58,18 @@ This controls whether a side bar application switcher ("Launcher" from Unity Des
 
 The arrays are rendered at the top (`head`), middle (`body`) and bottom (`tail`) of the sidebar. `head` and `tail` items remain static (pinned to the top and bottom respectively), while `body` items occupy the remaining space and scroll as a group when they overflow it. Each array is an ordered list of item IDs:
 
-| ID              | Description                                                       |
-| --------------- | ----------------------------------------------------------------- |
-| `running`       | Running applications; only valid in `body`, takes remaining space |
-| `magnifier`     | Magnifier toggle                                                  |
-| `cursor-scale`  | Cursor size cycle                                                 |
-| `output-filter` | Display colour filter cycle                                       |
+| ID              | Description                                              |
+| --------------- | -------------------------------------------------------- |
+| `running`       | Running applications (scrollable, takes remaining space) |
+| `magnifier`     | Magnifier toggle                                         |
+| `cursor-scale`  | Cursor size cycle                                        |
+| `output-filter` | Display colour filter cycle                              |
 
 ```{note}
-`running` is only supported in the `body` array, where it fills the remaining
-space and scrolls. A `running` entry placed in `head` or `tail` is ignored.
+`running` is best placed in `body`, where it fills the remaining space and
+scrolls. You can also place it in `head` or `tail`, but those sections are not
+height-constrained, so a long list of running apps may overflow the sidebar —
+use it there at your own risk.
 ```
 
 ```bash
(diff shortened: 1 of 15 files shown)


docs/reference/ubuntu-frame-configuration-options.md
```diff
@@ -55,17 +54,30 @@ This feature is only available on **Intel**, **AMD** and **ARM64** systems
...
+| `magnifier`     | Magnifier toggle                                         |
+| `cursor-scale`  | Cursor size cycle                                        |
+| `output-filter` | Display colour filter cycle                              |
+
+```{note}
+`running` is best placed in `body`, where it fills the remaining space and
+scrolls. You can also place it in `head` or `tail`, but those sections are not
+height-constrained, so a long list of running apps may overflow the sidebar —
+use it there at your own risk.
+```
```
tarek-y-ismail, line 73: Slightly unrelated: Is there any customer use case for `running` outside `body`? Maybe we should constrain it to there.

snap/hooks/configure
```diff
@@ -134,41 +134,38 @@ else
...
-    ;;
-  false|"")
-    launcher_enabled=false
-    ;;
   {*)
-    launcher_enabled=true
-    launcher_tail="$(snapctl get launcher.tail)"
-    if ! validate_launcher_tail "$launcher_tail"; then
-      echo "ERROR: launcher.tail must be a list of zero or more of: magnifier, cursor-scale, output-filter"
+    if ! launcher_items="$(printf '%s' "$launcher_val" | jq -er '
```
tarek-y-ismail: Wouldn't this be `true` if we pass an a launcher config object with empty head, body, and tail? Is `-e` the right tool?

tarek-y-ismail (commented):
Question, don't we want to do some sort of deduplication? In the example config you provide, `running` is seen twice, and it renders twice.

docs/reference/ubuntu-frame-configuration-options.md
saviq-work: Why?

saviq-work (commented):
> Question, don't we want to do some sort of deduplication? In the example config you provide, `running` is seen twice, and it renders twice.

PEBCAK

docs/reference/ubuntu-frame-configuration-options.md
tarek-y-ismail: I suppose it depends on how Frame is set up, but if the launcher is outside the body, it won't scroll, right?

saviq-work: So don't place it there if you risk it running out :)

tarek-y-ismail:
> Fixed in 77d2afd.

Did Copilot forget to push?

mattkae (approved):
This looks up to par now, nice ⛳

snap/hooks/configure
tarek-y-ismail: @copilot Take a look at this. It appears that earlier attempts at fixing this failed to push the changes.

tarek-y-ismail:
Looks like MSFT is either having infra trouble or didn't test Copilot with stacks...

snap/hooks/configure
```diff
@@ -134,41 +134,38 @@ else
...
-    launcher_tail="$(snapctl get launcher.tail)"
-    if ! validate_launcher_tail "$launcher_tail"; then
-      echo "ERROR: launcher.tail must be a list of zero or more of: magnifier, cursor-scale, output-filter"
+    if ! launcher_items="$(printf '%s' "$launcher_val" | jq -r '
+      if type != "object" then error("launcher must be an object")
+      elif (.head | type) != "array" then error("launcher.head must be an array")
+      elif ((.body // []) | type) != "array" then error("launcher.body must be an array")
+      elif (.tail | type) != "array" then error("launcher.tail must be an array")
+      else .head[], (.body // [])[], .tail[] end
+    ')"; then
```
saviq-work, line 148: @tarek-y-ismail it should actually be this, shouldn't it:
```suggestion
    if ! launcher_items="$(printf '%s' "$launcher_val" | jq -r '
      if type != "object" then error("launcher must be an object")
      elif (.head | type) != "array" then error("launcher.head must be an array")
      elif (.body | type) != "array" then error("launcher.body must be an array")
      elif (.tail | type) != "array" then error("launcher.tail must be an array")
      else .head[], .body[], .tail[] end
    ')"; then
```

We require all three to be arrays, otherwise `snapctl get launcher.body` will fail.

tarek-y-ismail: AFAIU, the `foo // []` syntax is just defaulting the value to an empty array if it's not specified, right? I suppose this is an improvement..

saviq-work: Yes, it is here, but not when you `snapctl get …` it by the full key.

tarek-y-ismail: Aha, I see, it would blow up then. Good catch.

tarek-y-ismail:
Copilot still failing to push...

tarek-y-ismail (approved):
Let's get it in!

PR #338, opened 2026-09-16
launcher: protect against bad config in the launcher

saviq-work:


tarek-y-ismail (commented):
Don't we wanna signal that the configuration is broken is some way? For example, showing a stub button with some icon?

saviq-work:
> Don't we wanna signal that the configuration is broken is some way? For example, showing a stub button with some icon?

Not sure, what could the user (think airport kiosk) do with that?

tarek-y-ismail:
I Should we blow up in debug and log in release then? Should let operators know something is up when testing and, assuming they build in release, should be fine in normal usage.

saviq-work:
> Should we blow up in debug and log in release then?

We never release anything in debug mode, no one would build this themselves. Operators should never really be able to get here, since the `configure` script validates the config input. And the failure mode is quite obvious - missing items that you've configured to be there.

launcher/lib/views/accessibility_button.dart
```diff
@@ -27,7 +27,12 @@ class AccessibilityButton extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final controller = GetIt.instance.get<AccessibilityController>();
-    final option = controller.options.firstWhere((o) => o.id == optionId);
+    AccessibilityOption option;
+    try {
+      option = controller.options.firstWhere((o) => o.id == optionId);
```
mattkae, line 32: Nit: I would just `return the widget` in the try and resolve it in the `option` field directly. Less code :)

mattkae (approved):
One suggestion, but looks good!

launcher/lib/views/accessibility_button.dart
saviq-work: @copilot apply the above suggestion

tarek-y-ismail: Looks like Copilot is still not cooperating?

tarek-y-ismail (approved):
I'll stop nagging :)

launcher/lib/views/accessibility_button.dart
saviq-work: Can you try? :)

tarek-y-ismail: @copilot+gpt-5.6-luna Would you kindly address this point of feedback and push the changes?

tarek-y-ismail: No dice.

saviq-work: @mattkae done :+1:.

